PR #1, opened 2019-07-30
CSB-507: implement basic roster models

waharnum:
To review this locally you'll need to run `python manage.py migrate` after doing the initial setup.

You should then be able (with the admin account created at initial setup) to:
- create Sites and Periods
- create Users with the additional `ClusiveUser` fields (at the bottom)


roster/models.py
```diff
@@ -0,0 +1,71 @@
+from django.db import models
+from uuid import uuid4
+from django.contrib.auth.models import User
+
+class Site(models.Model):
```
bgoldowsky: I like "site", but this is "Org" in OneRoster data model, so we should consider whether that's better.

roster/models.py
```diff
@@ -0,0 +1,71 @@
...
+    # at https://wiki.cast.org/display/CISL/User+model+development?
+    permission = models.BooleanField(default=False)        
+
+    periods = models.ManyToManyField(Period)
+
+    # TODO: consider creating a separate "Roles" model to store 
+    # roles, and refer to them with a Foreignkey - see comment 
+    # about "hacking choices to be dynamic" at https://docs.djangoproject.com/en/2.2/ref/models/fields/#choices
+    #
+
```
bgoldowsky: Also consider using Django "groups" for this.  Since these roles will mediate permissions to access various capabilities, might be a good fit.

roster/models.py
```diff
@@ -0,0 +1,71 @@
...
+class Site(models.Model):
+    name = models.CharField(max_length=100)
+    site_id = models.CharField(unique=True, default=uuid4, max_length=36)
+    # TODO: is this an address field? Should it be managed as such?
+    location = models.TextField(max_length=500)
+
+    # TODO: should these three be restricted by choices?
+    language_code = models.CharField(max_length=2, default='EN')
+    country_code = models.CharField(max_length=2, default='US')
+    timezone = models.CharField(max_length=30, default='EST')
```
bgoldowsky: timezone should use the official "tz" code, eg "America/New_York" rather than "EST".
https://en.wikipedia.org/wiki/List_of_tz_database_time_zones

roster/models.py
```diff
@@ -0,0 +1,71 @@
...
+    
+    # Django standard user class contains the following fields already
+    # - first_name
+    # - last_name
+    # - username
+    # - password
+    # - email
+    user = models.OneToOneField(User, on_delete=models.CASCADE)
+
+    subject_id = models.CharField(unique=True, default=uuid4, max_length=36)
```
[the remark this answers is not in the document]
waharnum: How have the anonymized IDs been assigned in the past? Some possibilities I can think of:

1. Method to create a human-readable but anonymized ID: (something like an autoincrementing `site-1` / `user-1` / `period-1`
2. `uuid`-style like what's done here
3. Manual assignment

roster/models.py
waharnum: I've now used `pytz` to generate this as a list of choices.

roster/models.py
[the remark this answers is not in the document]
bgoldowsky: Manual assignment (via spreadsheet import) has been our strategy.

roster/models.py
```diff
@@ -0,0 +1,82 @@
...
+available_timezones_friendly = []
+for tz in available_timezones:
+    tz_friendly = tz.replace("_", " ")
+    available_timezones_friendly.append(tz_friendly)
+
+class Site(models.Model):
+    name = models.CharField(max_length=100)
+    anon_id = models.CharField(unique=True, default=uuid4, max_length=36)
+    # TODO: is this an address field? Should it be managed as such?
+    location = models.TextField(max_length=500)
```
bgoldowsky: Change to city, state/province.

roster/models.py
```diff
@@ -0,0 +1,82 @@
...
+
+    # TODO: should these two be restricted by choices?
+    language_code = models.CharField(max_length=2, default='en')
```
bgoldowsky: Drop for now.

roster/models.py
```diff
@@ -0,0 +1,71 @@
...
+
+    # TODO: should these three be restricted by choices?
+    language_code = models.CharField(max_length=2, default='EN')
+    country_code = models.CharField(max_length=2, default='US')
+    timezone = models.CharField(max_length=30, default='EST')
+
+    def __str__(self):
+        return '%s (%s)' % (self.name, self.site_id)
+
+class Period(models.Model):
```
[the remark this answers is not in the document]
bgoldowsky: Keep Period.

roster/models.py
```diff
@@ -0,0 +1,82 @@
...
+
+    timezone = models.CharField(max_length=30, default='America/New York', choices=TIMEZONE_CHOICES)
+
+    def __str__(self):
+        return '%s (%s)' % (self.name, self.site_id)
+
+class Period(models.Model):
+    site = models.ForeignKey(Site, on_delete=models.CASCADE)
+    name = models.CharField(max_length=100)
+    anon_id = models.CharField(unique=True, default=uuid4, max_length=36)
```
bgoldowsky: Nullable, unique, default to null

roster/models.py
```diff
@@ -0,0 +1,82 @@
...
+    # - username
+    # - password
+    # - email
+    user = models.OneToOneField(User, on_delete=models.CASCADE)
+
+    anon_id = models.CharField(unique=True, default=uuid4, max_length=36)
+
+    # TODO: should this be an enum of states, as per comment 
+    # at https://wiki.cast.org/display/CISL/User+model+development?
+    permission = models.BooleanField(default=False)
```
bgoldowsky: Change to enumeration

roster/models.py
```diff
@@ -0,0 +1,82 @@
...
+    GUEST = 'GU'
+    STUDENT = 'ST'
+    TEACHER = 'TE'
+    RESEARCHER = 'RE'
+    ADMIN = 'AD'
+
+    ROLE_CHOICES = [
+        (GUEST, 'Guest'),
+        (STUDENT, 'Student'),
+        (TEACHER, 'Teacher'),
```
bgoldowsky: Add PARENT

PR #3, opened 2019-08-26
Feature/csb 510 events

bgoldowsky:
Alan, could you review this and offer any suggestions to improve it?


eventlog/models.py
```diff
@@ -0,0 +1,39 @@
+from django.db import models
+from uuid import uuid4
+from roster.models import Period, ClusiveUser, Roles
+from django.utils import timezone
+import caliper
+
+# A user session
+class Session(models.Model):
+    id = models.CharField(primary_key=True, default=uuid4, max_length=36)
+    user = models.ForeignKey(to=ClusiveUser, on_delete=models.CASCADE)
```
waharnum: We should confirm that the `on_delete` behaviour is what's desirable for all of the ForeignKey relationships, as there's been talk in the past of the issue of deleting a user but retaining anonymized data.

eventlog/tests.py
```diff
@@ -0,0 +1,71 @@
+from django.contrib.auth.models import User
+from django.test import TestCase
+from django.utils import timezone
+
+from roster.models import Site, Period, ClusiveUser
+from eventlog.models import Session, Event
+
+# TODO copied from roster tests. Is there a central place that some of this setup could be done?
```
waharnum: The `shared` directory that currently contains only the templates and static files could be used for this purpose. 

Alternately, it would be possible to refactor the Roster test functions so that they can suit both tests, then import the test functions from Roster like so: `from roster.tests import set_up_test_sites, set_up_test_periods, set_up_test_users` - this might be more appropriate, since these are related to creating test versions of the Roster model objects.

eventlog/models.py
```diff
@@ -0,0 +1,39 @@
+from django.db import models
+from uuid import uuid4
+from roster.models import Period, ClusiveUser, Roles
+from django.utils import timezone
+import caliper
+
+# A user session
+class Session(models.Model):
```
waharnum: I think we should name this something like `ClusiveSession`, `SessionEvent` or similar to avoid possible confusion / collision with the `Session` object in `django.contrib.sessions.models`.

eventlog/signals.py
```diff
@@ -0,0 +1,58 @@
...
+        user_agent = kwargs['request'].META.get('HTTP_USER_AGENT', '')
+        session = Session(user=clusive_user, userAgent=user_agent)
+        session.save()
+        # Put the ID of the database object into the HTTP session so that we know which one to close later.
+        http_session = kwargs['request'].session
+        http_session['db_session_id'] = session.id.__str__()
+        # Store the user's "Current Period" - in the case of teachers who may be associated with more than one Period,
+        # this will be the first-listed one.  Teachers should have some mechanism of changing their current period
+        # in the user interface, so they can interact with any of their classes; this should update the current period
+        # as stored in the session so that event logging will use the correct one.
```
waharnum: Not necessarily for this round, but this comment makes me think we may want to revisit the data model of how sessions and users are related if there are significant differences between how students and teachers relate to them.

eventlog/signals.py
```diff
@@ -0,0 +1,58 @@
...
+        event = Event(type='SESSION_EVENT',
+                      action='LOGGED_IN',
+                      actor=clusive_user,
+                      session=session,
+                      group=Period.objects.get(id=current_period_id) if current_period_id else None,
+                      membership=clusive_user.role,
+                      )
+        event.save()
+    except ClusiveUser.DoesNotExist:
+        print("Login by a non-Clusive user")
```
waharnum: Longer term we should log at Django's logging classes rather than `print` (my own code does this as well right now): https://docs.djangoproject.com/en/2.2/topics/logging/

eventlog/tests.py
```diff
@@ -0,0 +1,71 @@
...
+class EventlogTestCase(TestCase):
+    def setUp(self):
+        set_up_test_sites()
+        set_up_test_periods()
+        set_up_test_users()
+
+    def test_session_created(self):
+        login = self.client.login(username='user1', password='password1')
+        self.assertTrue(login)
+        session = Session.objects.all()[0]
```
waharnum: Is there a more reliable way to do this (get the session associated with the login)? The assumption that the first object will be the associated session may be true in the testing context right now, but I worry about possible issues further down the line. I'd suggest doing something like sorting the queryset by the session creation date and getting the latest if there's no way to retrieve the associated session event based on information from the login.

I would also use `all().first()` in place of `all()[0]` when doing these kind of queries, as described at https://docs.djangoproject.com/en/2.2/ref/models/querysets/#first

eventlog/tests.py
```diff
@@ -0,0 +1,71 @@
...
+        self.assertEquals(session.user, ClusiveUser.objects.all()[0], "Session should connect to the user")
+        self.assertEquals(session.startedAtTime.date(), timezone.now().date(), "Should set reasonable session start time")
+        self.assertIsNotNone(session.userAgent)
+
+    def test_session_closed(self):
+        login = self.client.login(username='user1', password='password1')
+        self.assertTrue(login)
+        self.client.logout()
+        session = Session.objects.all()[0]
+        self.assertEquals(session.endedAtTime.date(), timezone.now().date(), "Should set reasonable session end time")
```
waharnum: The default session timeout looks to be two weeks: https://docs.djangoproject.com/en/dev/ref/settings/#session-cookie-age - we should be able to adjust this in `settings.py`

It occurs to me that we should also verify that a session timeout sends the same signal as a manual logout, if we want to be able to record it as one.

eventlog/models.py
bgoldowsky: Setting to PROTECT for now. That should prevent deletes of users, sessions, groups if they have event data attached to them.  We can figure out alternative methods later, but this should keep us safe for now.

eventlog/tests.py
bgoldowsky: Will import from roster.tests.

eventlog/models.py
bgoldowsky: LoginSession is what we've used in previous projects, will change to that.

eventlog/signals.py
bgoldowsky: Changed this to log.warning and out in a basic logging config

eventlog/tests.py
bgoldowsky: Changed to querying for most recent loginsession.

eventlog/tests.py
bgoldowsky: By default Django doesn't seem to have a notion of a session timing out.
May need add on such as https://pypi.org/project/django-session-idle-timeout/ ?

PR #21, opened 2019-12-06
CSB-100: Add night theme and *hack* in some color updates to reader frame

mbrambilla:
The root page should be mostly up to date the the latest round of colors from design.  No idea if they have been updated in JIRA.

I am sure there is a better way to get the color overrides into the reader frame, but they are mostly there.  So many conflicting rules and specificity issues to get them all working in short order.

Additional note: the over reliance on `!important` in the CSS rules for Readium, UIO, and everything else is going to be a problem.


mbrambilla:
Also adjusted the image blend modes to fix an issue with the images becoming too dark in night mode.

PR #23, opened 2020-02-18
CSB-782: TTS selection

waharnum:
This improves the TTS to accomodate reading selected text only. Also includes some high-level functionality for dealing with selection text such as determining direction.


bgoldowsky:
This looks good to me!

PR #24, opened 2020-02-20
CSB-828: Fix draggable glossary popover behavior - updates to Figurat…

mbrambilla:
…ion and Popper

Creating this as a PR since the method for including Popper into the Webpack needed to change in order for it to be included.  Not sure if this is the correct way, so looking to get a second opinion on it.

- Made upstream fixes and pushed a new Figuration release to deal with `fixed` position draggable popovers.  Also updated the Popper.js dependency to latest v1.x version.  
- No plans to update to Popper v2.x at this time.  The 2.x version drops support for IE11, which Figuration v4.x supports. Would also require significant rework in order to update.


bgoldowsky:
This is working for me. But why was the webpack change needed?

mbrambilla:
For some reason webpack would package Popper v1.15.0 into main.js. But, would not do so with v1.16.x using the same method.  Not sure why, have not looked into the changelog for Popper.

Found a [question on Stack Overflow](https://stackoverflow.com/q/45645971) that resulted in the current implementation.

PR #27, opened 2020-04-27
CSL-119: Store user settings in database / restore on login

waharnum:
MVP for saving and restoring preferences from the database on a per-user basis. JIRA task: https://castudl.atlassian.net/browse/CSL-119?atlOrigin=eyJpIjoiMmU0MmNkN2NiYmJmNDMxNGIyYzMzZDkyM2VkMGE2MGIiLCJwIjoiaiJ9


src/eventlog/admin.py
```diff
@@ -1,17 +1,28 @@
...
     readonly_fields = ('id', 'eventTime', 'actor', 'group', 'membership', 'type', 'action', 'session',
                        'document', 'page', 'control', 'value')
     list_display = ('id', 'eventTime', 'actor', 'type', 'action')
     ordering = ('-eventTime',)
 
+
+class PreferenceAdmin(admin.ModelAdmin):
+    readonly_fields = ('id',)
+    list_display = ('id', 'user', 'pref', 'value')
+    ordering = ('user', 'pref')
```
bgoldowsky, line 23: Shouldn't this be in roster/admin.py ?

waharnum: On the face of it, I agree - it looks like it's from your work to add the preference saving here: https://github.com/cast-org/clusive/blob/feature/CSB-520-save-preferences/src/eventlog/admin.py - did it need to be in here to administer the preference events?

PR #29, opened 2020-05-06
CSL-187: No-op console.log() when not in debug mode

mbrambilla:
I know Alan has some ideas on how to handle reducing the console output, but wanted to toss this out there.

This might work to reduce most of the console output from both the Fluid and JS sides.

Concept uses Django's `DEBUG` setting as master setting to enable/disable.
This no-op method (while potentially inefficient) allows the console.log to be re-enabled in production if some remote debugging is needed.

~Still need to investigate how to pass this into the Readium frame.~

Thoughts?

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -7,5 +7,6 @@ nltk==3.4.5
 lemminflect==0.2.0
 future==0.18.2
 numpy==1.17.4
+django-settings-export==1.2.1
 -e git+https://github.com/IMSGlobal/caliper-python@1.1.0.2#egg=imsglobal_caliper
 -e git+https://github.com/cast-org/django-session-timeout#egg=django-session-timeout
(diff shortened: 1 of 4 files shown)


waharnum:
My thoughts:

1. I think using the Django `DEBUG` setting to manage the log level is a great idea for making front-end code issues easier to debug in production or other "live" environments if necessary.

2. If we want to no-op for now to avoid spamming the console in production I'd say go ahead; we can replace it when we've got a little more time to implement something better.

mbrambilla:
> I think that we should make sure this doesn't have an unexpected result for messages to the console (like from browser debugging tools) that our application doesn't produce.

As implemented in this PR, it noops all `console.log` calls, so every call to this method gets 'blocked'.

We could also implement our own log function with a level system to use some of the [more common console methods](https://developer.mozilla.org/en-US/docs/Web/API/Console#Usage).  It looks like there is [similar Infusion log handling](https://docs.fluidproject.org/infusion/development/CoreAPI.html#logging-and-error-handling) , but I could not seem to get it to work just from the browser console. Probably missed something on my end.

Commonly used methods all appear to be supported in the target browsers:
 - `console.log()`
 - `console.info()`
 - `console.warn()`
 - `console.error()`

PR #30, opened 2020-05-12
CSL-236: fix glossary on multi-page preferences

waharnum:
Created as https://castudl.atlassian.net/browse/CSL-236 for effort expended. This also fixes an issue that was occurring on my local environment with SQLite with the glossary entries in large texts.


bgoldowsky:
Works as long as the 'glossary' preference has been set to a true value, not when it's null but glossary is on by default.
However, this should be a moot point shortly when preference defaults are set on the server side, so I think we can call this done and merge it.

PR #32, opened 2020-06-03
CSB-293 & others (server-side preference storage features and improvements)

waharnum:
This should address several outstanding JIRA issues related to the server-side preference storage; they were all somewhat entangled or related, so I've done them all as part of this PR:

* `CSL-185: Remove cookie usage in get/set of Preference Store`
* `CSL-186: Store default preference state on server side`
* `CSL-293: Refactor Django-side preference storage API`
* `CSL-108: Change default line spacing to 'middle' option`

**Notes**
Switching to different stored preference sets (as part of `CSL-186`) have been implemented in a quite lightweight way); I've created a new model in `Roster` for them and loaded the defaults using Django's `fixtures` features, as described at https://docs.djangoproject.com/en/3.0/howto/initial-data/

This means an additional step is required when doing the initial DB setup:  `python manage.py loaddata prefencesets` (I've added it to the `INSTALL.md` document) - we may want to consider another approach longer term that stores the preference sets as pure JSON files, but this was a very expedient way to do it right now.


src/roster/tests.py
```diff
@@ -206,28 +210,59 @@ def test_permissioned_property(self):
...
+        handler.assert_called_once()
+        # TODO: clarify what this is doing; is it intended to test the handler?        
+        handler.calls.kwargs.preference.pref='baz'
+        handler.calls.kwargs.preference.value='lur'
 
 class PageTestCases(TestCase):
 
         def setUp(self):
-            set_up_test_users()
+            set_up_test_users()
```
bgoldowsky, line 265: Thank you for adding tests!

src/roster/views.py
```diff
@@ -20,28 +22,89 @@ def guest_login(request):
...
+
+        user = ClusiveUser.from_request(request)        
+        set_user_preferences(user, request_prefs, request)
+        for req_pref_key in request_prefs:
+            req_pref_val = request_prefs[req_pref_key]
+            preference = user.get_preference(req_pref_key)               
+            if(preference.value != req_pref_val):                           
+                preference.value = req_pref_val
+                preference.save()
+                preference_changed.send(sender=ClusiveUser.__class__, request=request, preference=preference)
```
bgoldowsky, line 45: Is this redundant?  Looks like it's the same as what set_user_preferences does.

src/frontend/js/clusive-prefs-panel.js
```diff
@@ -71,32 +71,32 @@
...
+            var setURL = directModel.setURL;            
+            $.ajax({
+                type: "POST",
+                url: setURL,
+                headers: {
+                    'X-CSRFToken': DJANGO_CSRF_TOKEN
+                },
+                data: JSON.stringify(fluid.get(model, 'preferences')),
+                success: function (data) {
+                    console.log("set preferences", data);
```
bgoldowsky, line 96: TODO: handle failure case

INSTALL.md
```diff
@@ -70,6 +70,7 @@ The local development configuration uses sqlite3, so no database setup is requir
 
 Run in the /target directory:
 * `python manage.py migrate`
+* `python manage.py loaddata prefencesets`
```
bgoldowsky, line 73: Should this be added to entrypoint.sh for docker startup as well?

src/roster/fixtures/preferencesets.json
```diff
@@ -0,0 +1,11 @@
+[
+    {
+        "model": "roster.preferenceset",
+        "pk": 1,
+        "fields": {
+            "set_name": "default",
+            "set_description": "Default Clusive preference set",
+            "prefs_json": "{\"fluid_prefs_letterSpace\":1,\"fluid_prefs_contrast\":\"default\",\"theme\":\"default\",\"fluid_prefs_textFont\":\"default\",\"textFont\":\"default\",\"fluid_prefs_textSize\":1,\"textSize\":1,\"fluid_prefs_lineSpace\":1.6,\"lineSpace\":1.6,\"cisl_prefs_glossary\":true}"
```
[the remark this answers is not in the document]
waharnum: We can safely remove the `fluid_prefs_*` ones from the file; done.

src/roster/views.py
waharnum: Yes, I cleverly factored that out into a function, then forgot to actually remove it. Fixed.

INSTALL.md
waharnum: Added; I wasn't sure of the right point to add it to the container build.

src/frontend/js/clusive-prefs-panel.js
```diff
@@ -71,32 +71,32 @@
...
-            });
+            $.ajax({
+                type: "POST",
+                url: resetURL,
+                headers: {
+                    'X-CSRFToken': DJANGO_CSRF_TOKEN
+                },
+                data: JSON.stringify({adopt: 'default'}),
+                success: function (data) {
+                    console.log("reset preferences to default", data);
```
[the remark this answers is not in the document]
waharnum: This was happening by "coincidence" (reset default on client is same as reset default from server side); I've now made it explicit. It could be improved to reduce the need to hit the server twice; I'll create a JIRA for this.

src/frontend/js/clusive-prefs-panel.js
waharnum: Added basic logging for failure.

waharnum:
@bgoldowsky I think all comments should be addressed now (and the code much improved for doing so).

PR #37, opened 2020-07-22
Add message queue MVP

waharnum:
Server-side work is still needed to implement signals and persistent messages to be processed, but I'm opening this PR as an MVP and will create some new JIRA issues to handle further work.

Notes:
* Client-side tests are in `/src/frontend/tests/messageQueue/index.html` - I ran these using a local static server (`http-server`) during development
* The integration with the preference store code is in `/src/frontend/js/clusive-prefs-django-store.js`
* `src/messagequeue/views.py` is the current handler for processing preference changes on the server side, which will be rewritten to use signals


src/frontend/js/clusive-prefs-django-store.js
```diff
@@ -0,0 +1,167 @@
...
+            
+        } else {
+            var getURL = directModel.getURL;
+        
+            $.get(getURL, function(data) {
+                console.debug('Get user preferences from the server ', getURL);
+                console.debug('Received preferences: ', data);
+                djangoStorePromise.resolve({
+                    preferences: data
+                });
```
bgoldowsky, line 118: Looks like this doesn't do any caching of response, so will not resolve the issue of multiple server calls to get preferences. That will be a separate fix later?

src/frontend/js/clusive-message-queue.js
```diff
@@ -0,0 +1,144 @@
...
+            },
+            "onCreate.setFlushInterval": {
+                funcName: "{that}.setFlushInterval"                
+            },
+            "queueShouldFlush.flushQueue": {
+                funcName: "{that}.flush"
+            },
+            "queueFlushSuccess.clearQueue": {
+                func: "{that}.clearQueue",                
+            }
```
bgoldowsky, line 36: If I'm understanding correctly, the sequence is:

1. Timer invokes queue flush
2. Current message list read from localstorage; network connection to server initiated
3. After server round trip finishes, clearQueue is invoked
4. Content of local storage is erased.

What happens if between steps 2 and 3 (during the time we're waiting for a response from the server, which could be a few seconds) new messages are added to the queue? Wouldn't they be lost?

src/messagequeue/models.py
```diff
@@ -0,0 +1,15 @@
...
+    PREF_CHANGE = 'PC'
+
+    CHOICES = [
+        (PREF_CHANGE, 'Preference Change'),
+    ]    
+
+class Message(models.Model):
+    type = models.CharField(max_length=10, choices=MessageTypes.CHOICES)
+    timestamp = models.DateField()
+    content = models.TextField(max_length=500)
```
bgoldowsky, line 15: Is this a permanent new thing or a temporary thing for testing?

src/messagequeue/views.py
```diff
@@ -0,0 +1,29 @@
...
+        if(message["content"]["type"] == 'PC'):
+            print("Found a preference change message")
+            print(message)
+            set_user_preferences(user, message["content"]["preferences"], request)
+
+class MessageQueueView(View):
+    def post(self, request):        
+        try:
+            messages = json.loads(request.body)        
+            user = ClusiveUser.from_request(request)
```
bgoldowsky, line 25: With latest code you can just do request.clusive_user

bgoldowsky:
Just remembered, we also talked about having the "flush" operation send the client's current time, so that the server could calculate and apply an appropriate adjustment to the timestamps.

src/frontend/js/clusive-message-queue.js
waharnum: Good point - writing some code to address this right now and will write a test that simulates a non-instant round trip to make sure this is addressed and there's not a chance of message loss.

src/messagequeue/models.py
waharnum: This might be unnecessary unless we want to be able to hold messages in the database for asynchronous processing. My thoughts were going somewhere like this (though the implementation is not finished):

- create a Message model
- use the Signals to have relevant areas of code listen for the creation of Message objects relevant to them (via the `type` field)
- have the relevant area of code responsible for deleting a message once it has been properly acted on

waharnum:
> 
> 
> Just remembered, we also talked about having the "flush" operation send the client's current time, so that the server could calculate and apply an appropriate adjustment to the timestamps.

I'll add this - all the messages are individually timestamped right now but this isn't adequate for applying an adjustment between the server time and the client reported time.

src/frontend/js/clusive-prefs-django-store.js
waharnum: I'll add a debounce function here.

waharnum:
@bgoldowsky I think this should now address all the comments above - I'm going to make a separate JIRA for completing the signals-based implementation that will improve the decoupling and make this more generally useful for other uses than preference changes.

waharnum:
Closed, work has been superseded by this PR: https://github.com/cast-org/clusive/pull/46

PR #46, opened 2020-08-18
CSL-404: signal-based message queues

waharnum:
This adds signal-based message queue code and a number of other changes to support future work.


src/frontend/js/clusive-prefs-django-store.js
```diff
@@ -0,0 +1,188 @@
...
+            var getURL = directModel.getURL;
+            
+            // Use cached result if within debounce time
+            if(timeDiff < debounce && that.lastResponse) {
+                console.debug('Get user preferences from cache of last result');
+                djangoStorePromise.resolve({
+                    preferences: that.lastResponse
+                });
+            } else if(that.requestIsInFlight) {
+                djangoStorePromise.reject('Won\'t get user preferences from server, another GET request is currently in flight')
```
bgoldowsky, line 139: I am seeing this message in the console when trying this out.  Is that expected?
Not clear that it resolves the request at all in this case.

src/shared/templates/shared/partial/head_scripts.html
```diff
@@ -7,14 +7,22 @@
...
 // Initialized with the results of createClusivePrefsEditor() when that's called
 var clusivePrefs;
 </script>
 
 <script src="{% static 'shared/js/lib/main.js' %}"></script>
 <script src="{% static 'shared/js/clusive-reader-prefs.js' %}"></script>
+<script src="{% static 'shared/js/clusive-prefs-django-store.js' %}"></script>
 <script src="{% static 'shared/js/clusive-prefs-panel.js' %}"></script>
+<script src="{% static 'shared/js/clusive-prefs-modal-settings.js' %}"></script>
+<script src="{% static 'shared/js/clusive-message-queue.js' %}"></script>
```
bgoldowsky, line 25: We should perhaps update the build to bundle more of these into main.js, unless there's a reason not to.

src/messagequeue/migrations/0001_initial.py
```diff
@@ -0,0 +1,23 @@
+# Generated by Django 2.2.13 on 2020-07-20 14:08
```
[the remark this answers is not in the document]
bgoldowsky: I squashed them.

PR #47, opened 2020-09-01
CSL-405 and CSL-406

waharnum:
Improvements to the message queue implementation:

* CSL-405: flushing of message queue on logout / prompt using `beforeunload` if queue not flushed
* CSL-406: filtering of messages by user ID, https://castudl.atlassian.net/browse/CSL-406


bgoldowsky:
I tried it out, seems to work well!

PR #50, opened 2020-09-04
CSL-297: preference reset for message queue implementation

waharnum:
* Reset now working with new message queue implementation
* Remove the `beforeunload` event until it can be made to work only on logout (see https://castudl.atlassian.net/browse/CSL-463)


bgoldowsky:
Looks good.

PR #51, opened 2020-09-04
CSL-455:  TTS control update

mbrambilla:
Adds state swapping for the TTS button in the sidebar, along with Pause and Resume controls.  Works for the most part with the ClusiveTTS, but there does seem to be some lag when using pause/resume, but I don't think there is much we can do there.

Tried to build in support multiple regions of TTS controls, such as a preview one in the settings modal. Does not take into account reading selected text from context menu since it is Reader controlled.

Probably going to need some form of callback from the Reader to update state when play is started or stopped.

### DIFF
--- a/src/frontend/html/library.html
+++ b/src/frontend/html/library.html
@@ -92,7 +92,7 @@
                 </div>
             </div>
 
-            <div class="sidebar sidebar-end" role="region" aria-label="Reading tools">
+            <div class="sidebar sidebar-end">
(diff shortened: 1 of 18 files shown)


mbrambilla:
This also has some dependencies on PR #49 in terms of some icons and CSS classes.

mbrambilla:
Argh , merged in the wrong direction. Hopefully not too breaking.

PR #52, opened 2020-09-15
CSL-463: prompt to save on logout

waharnum:
JIRA: https://castudl.atlassian.net/browse/CSL-463

This programatically attaches and removes an `beforeUnload` event when interaction is detected with the logout prompt.

### DIFF
--- a/src/frontend/js/clusive-prefs-django-store.js
+++ b/src/frontend/js/clusive-prefs-django-store.js
@@ -48,7 +48,7 @@
         $(logoutLinkSelector).mouseenter(
             function () {   
                 if(! that.isQueueEmpty()) {
-                    clusive.djangoMessageQueue.addUnloadListener(that);
+                    addUnloadListener(that);
                     console.debug("mouseenter event on logout link, flushing message queue.");
                     that.flush();        
                 }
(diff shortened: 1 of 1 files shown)


mbrambilla:
There is the minor edge case of the logout having being in both the hover and focus states at the same time.  It looks like if either of those states is removed, `removeEventListener()` is executed without checking the other state has also been removed.

bgoldowsky:
I think the user experience of this is less than ideal.  If you logout at certain times you'll get a warning that there is unsaved data, but have no sense of how to fix that since there's no save button.  Then you'll try again and it will mysteriously work.

How about an alternate strategy: 
-  Make a "logging out" page that just has the text "logging out" on it.  
- Have the logout link send you there.
- Javascript on that page will initiate a save (if one is needed), and when that is complete, log you out immediately.

waharnum:
> 
> 
> I think the user experience of this is less than ideal. If you logout at certain times you'll get a warning that there is unsaved data, but have no sense of how to fix that since there's no save button. Then you'll try again and it will mysteriously work.
> 
> How about an alternate strategy:
> 
>     * Make a "logging out" page that just has the text "logging out" on it.
> 
>     * Have the logout link send you there.
> 
>     * Javascript on that page will initiate a save (if one is needed), and when that is complete, log you out immediately.

This seems reasonable - if we are relying on Javascript we may be able to do this flow without a separate page by intercepting the link-follow event, canceling it, then sending the browser to the logout page once the save callback has completed. Let me look at some options, including what to do if the save fails or takes a long time to return.

waharnum:
Closed, superseded by https://github.com/cast-org/clusive/pull/53

PR #53, opened 2020-09-24
CSL-504: save on logout

waharnum:
Replaces the prior save-on-logout implementation.

1. Intercepts a logout click if the message queue is not empty.
2. Attempts to flush the messages.
3. Completes the logout after the flush attempt finishes (whether or not it was successful).


bgoldowsky:
I think this is a good improvement.  I noticed two things in testing:

- The "Saving changes..." message is not visible since the pulldown menu containing the logout link closes.
- If the user session times out, and then preferences are changed on the stale page, things can get in a weird state.

I think we can make separate JIRA issues for those, though, and merge this.

waharnum:
> 
> 
> I think this is a good improvement. I noticed two things in testing:
> 
>     * The "Saving changes..." message is not visible since the pulldown menu containing the logout link closes.
> 
>     * If the user session times out, and then preferences are changed on the stale page, things can get in a weird state.
> 
> 
> I think we can make separate JIRA issues for those, though, and merge this.

Created JIRAs for these two:
- https://castudl.atlassian.net/browse/CSL-509 ("Saving changes" message)
- https://castudl.atlassian.net/browse/CSL-510 ("weird state" situation - **Question**: what's the easiest way to replicate a session time out in order to fix this?)

PR #54, opened 2020-10-05
CSL-309: Update line height

mbrambilla:
Normalizes the `line-height` (line spacing) visuals between the Clusive and Readium sides.

Internal names for the line-height have been adjusted to align the `1.6` option as `default`, changing from the `1.2` option marked as such.  I don't think additional changes are needed since `1.6` is the default setting, unless there is a test that needs to be updated.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2735,9 +2735,9 @@
       "dev": true
     },
     "figuration": {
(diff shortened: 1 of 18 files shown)


bgoldowsky:
This looks good to me.

PR #55, opened 2020-11-05
NOJIRA: Add alternative method for TTS highlighting

mbrambilla:
This only affects the Clusive side of TTS, and most likely only for iOS.

This method uses a regex match to find the word boundary and start of next word to determine highlight.

Android/Chrome does not seem to support `utternance.onboundary` at the moment.
https://bugs.chromium.org/p/chromium/issues/detail?id=521666


bgoldowsky:
Looks good to me!

PR #60, opened 2020-11-16
Task - add book location information to pref change & tool use events

waharnum:
Adds information from the client-side Reader about the current location in the book to the preference change and control used events: `href` (current page in a multi-page book) and 'progression'.


bgoldowsky:
I thought progressions were numeric values - would it make sense to store as float rather than string?

waharnum:
> 
> 
> I thought progressions were numeric values - would it make sense to store as float rather than string?

Yes, this makes sense - done.

PR #67, opened 2021-01-29
CSL-708: debugging tools for message queue

waharnum:
* Adds a debug page at `/debug`
* Various changes to the message queue code to support better debugging


bgoldowsky:
I’m getting Uncaught TypeError: currentLogValue is null in the console (on Firefox). Page is not updating when queue is flushed.

waharnum:
> 
> 
> I’m getting Uncaught TypeError: currentLogValue is null in the console (on Firefox). Page is not updating when queue is flushed.

Sorry, I missed this comment on Friday - this should be fixed now (wasn't properly guarding against null values)

PR #68, opened 2021-02-03
bugfix/CSL-721: not all queues being flushed on logout as intended

waharnum:
This fixes the bug with minimal changes to the existing code:
- adds a singleton component (`logoutFlushManager`) to track the number of flushes at logout and compare them against the number of `djangoMessageQueues`, which are now registered with the manager at creation
- do not perform the logout until the expected number of queues have been flushed (the bug was because there are now two separate message queues, and whichever one called back first caused the logout before the other could fush)

Longer-term, I'd like to move all "flush on logout" behaviour into the manager component, but that can be a separate JIRA.

### DIFF
--- a/src/frontend/js/clusive-django-message-queue.js
+++ b/src/frontend/js/clusive-django-message-queue.js
@@ -3,6 +3,17 @@
 (function(fluid) {
     'use strict';
 
+    fluid.defaults("clusive.logoutFlushRecorder", {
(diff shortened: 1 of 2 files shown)


waharnum:
@bgoldowsky - please review when you have a chance. The easiest way to see the bug is fixed is to check the current behaviour in `development` in the debug page (one queue always errors when logging out) vs. this branch, which should fix the issue.

PR #69, opened 2021-02-05
Bug fix: CSL-712, glossary not always indicated

waharnum:
PR for fixing issues relating to the glossary not always being indicated. Currently in draft because there may be multiple issues.

### DIFF
--- a/src/frontend/js/clusive-prefs-django-store.js
+++ b/src/frontend/js/clusive-prefs-django-store.js
@@ -109,16 +109,17 @@
             
             // Use cached result if within debounce time
             if(timeDiff < debounce && that.lastResponse) {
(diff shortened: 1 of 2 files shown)


waharnum:
Issue fixed with `d02eda4` commit:
- user preferences were not always saved on the server-side when manually adjusting them back to their defaults
- this was most visible with the glossary because it's a toggle, but was an issue affecting other preferences as well

Commit ensures that all current client-side preferences are sent to the server, including those in the default state (server already ensures that values that haven't changed aren't saved)

waharnum:
@bgoldowsky I think this PR should fix the two major causes identified so far; if the problem continues to recur we'll need to further investigate.

PR #73, opened 2021-03-08
Feature/csl 749 Google single sign on

klown:
Code changes to allow single sign on using a user's Google credentials/account.  This is still a work in progress:
- need to remove a number of `django-allauth` endpoints as superfluous to Clusive, e.g.,  their "Sign-up" endpoint (`/accounts/signup`)
- likely need to adjust `requirements.txt` to include `django-allauth`

@bgoldowsky @waharnum While I've marked this as a draft pull request (see list above), there is enough here to start reviewing it.


klown:
This has been replaced with https://github.com/cast-org/clusive/pull/75.

Closing.

PR #79, opened 2021-04-13
CSL-864: comp check

waharnum:
Implements the MVP comprehension check.


src/shared/static/shared/font/fontello/config.json
```diff
@@ -631,6 +631,8 @@
       "search": [
         "thought"
       ]
+<<<<<<< HEAD
+=======
```
bgoldowsky, line 635: Looks like merge had problems. I think if you copy the fontello directory from current development it should work.

bgoldowsky (changes requested):
There were Fontello merge problems

src/shared/static/shared/font/fontello/config.json
waharnum: Sorry, thought I got all of these - too early in the morning to do merges I guess. Fixing now.

PR #81, opened 2021-04-21
CSL-917: generate robots.txt file blocking /account

waharnum:
This should block crawlers that respect `robots.txt` from following any `/account` URLs, including the guest login.

Found at https://stackoverflow.com/questions/18424260/django-serving-robots-txt-efficiently, which suggests serving `robots.txt` as a static file, but I'm uncertain of our static file server configuration in production, and this should be fine for now given our traffic levels.


waharnum:
@bgoldowsky, let me know if this looks OK for now and we can merge.

PR #84, opened 2021-04-28
CSL-945: add build scripts

waharnum:
Adds build scripts to automate away (most) of the steps for setting up or refreshing a local dev environment.

I've tested these myself (using WSL to test the `.sh` on my Windows machine), but if someone using a Mac can confirm the `.sh` script works for them, I think we can merge this and make people's lives a bit easier.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -6,7 +6,7 @@ To try it out quickly using Docker skip ahead to that section.
 
 ## Local Installation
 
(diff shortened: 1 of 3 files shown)


bgoldowsky:
Mostly worked on my Mac.

The line: 
   ./env/scripts/activate
needs to be
   ./env/bin/activate
for my system.

Also it requires that the Python virtualenv has been set up, with the correct version of python, under a subdirectory called "env".  This could be made a little more clear.

klown:
Would it be useful if this was another grunt task?  Actually, it couldn't be one task, but a target for Mac and another for Windows, e.g., `grunt macInstall` vs. `grunt winInstall`.

PR #92, opened 2021-05-28
CSL-971:  fix inconsistent theme application

waharnum:
Updated comment: with the 1.8.3-beta.1 version of the Reader, this problem seems to be entirely resolved. If @bgoldowsky can test, I think we can merge this one (perhaps after updating the dependency to the prod version Aferdita is going to publish)

- bumps Reader version to 1.8.3-beta.1
- uses the `resourceReady` API of the Reader before calling code that needs to interact with the Reader.

### DIFF
--- a/src/pages/templates/pages/reader.html
+++ b/src/pages/templates/pages/reader.html
@@ -205,11 +205,9 @@ var manifest_path = "{{ manifest_path }}";
                     console.debug("D2Reader finished loading the resource at: " + new Date());
                     $(document).ready(function () {
                         buildTableOfContents();
-                        buildAnnotationList();      
-                        setTimeout(function () {
-                            console.debug("Creating prefs editor at: " + new Date());
-                            clusivePrefs = createClusivePrefsEditor();
-                        }, 500);                                                        
+                        buildAnnotationList();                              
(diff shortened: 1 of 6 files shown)


src/pages/templates/pages/reader.html
```diff
@@ -201,6 +201,17 @@ <h2 class="modal-title">Image Details</h2>
...
             api: {
+                resourceReady: function() {
+                    console.debug("D2Reader finished loading the resource at: " + new Date());
+                    $(document).ready(function () {
+                        buildTableOfContents();
+                        buildAnnotationList();   
+                        // We only need to create the prefs editor once for a page
+                        if(typeof clusivePrefs !== undefined) {
+                            clusivePrefs = createClusivePrefsEditor();                        
+                        }
```
bgoldowsky, line 212: This feels backwards - don't you want to create it if the variable IS undefined, and leave it alone if it's already set?

Also, feels like building the TOC and annotation list also only need to be done once.

waharnum: Yes, test was backwards; now fixed.

I've created a `firstResourceLoaded` boolean flag we can use to do things only once on the page when a resource is first completely loaded by the Reader, rather than every time a new resource is loaded within the document.

PR #111, opened 2021-07-19
CSL-1092: page timing event tracking when using browser-cached page

waharnum:
Opening PR for discussion in tech meeting tomorrow, I think this is a simple possible solution to https://castudl.atlassian.net/browse/CSL-1092, but it needs to be validated as acceptable from a research statistics standpoint.

The high level change is that Page Timing events (sent from the client-side message queue system) now add their time to the `duration` and `active_duration` fields of an existing Event, rather than simply being rejected. The `load_time` field remains fixed to capture the initial render time of the page from the server.

Functionally, this means using the 'back' button or other means to return to a page via the browser history will generate additional duration and active duration for the initial page view event, but not a second page view event.

Questions:
- is this acceptable from a research statistics standpoint? 
- if it isn't, what do we need to capture that is missing? if we want to count going back through the history as a separate "view", could we add an increment on the original Event or similar as part of the Page Timing message?
- any potential side effects from this change that would be undesirable to research statistics or otherwise?

### DIFF
--- a/src/eventlog/signals.py
+++ b/src/eventlog/signals.py
@@ -38,45 +38,17 @@ def log_page_timing(sender, **kwargs):
     Collects and stores page timing information, which the client may send at the end of a page view.
     The extra information will be added in to the original VIEWED event representing the page view.
     """
-
-    # update the active time of relevant paradata based on a supplied event and 
-    # time
-    def update_active_times(event, added_time):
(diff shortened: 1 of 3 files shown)


waharnum:
I've reverted the prior approach and implemented the `never_cache` decorator on the `EventMixin` class, as discussed in today's tech meeting.

PR #113, opened 2021-07-27
CSL-1034: new CallToAction framework.

bgoldowsky:
Defines CallToAction model, similar to onboarding tips in concept.
There's a table of CTAs and another table to track user history of each.
Enabled 3 CTAs on the dashboard: welcome panel, summer reading (guest), summer reading (student).
Star rating will be moved into this framework but that hasn't yet been done.
Demographic survey will be added as well, stubbed in for now.

Note you'll need to apply migration and load the data fixture.


src/tips/models.py
```diff
@@ -100,3 +107,135 @@ def available_tips(cls, user: ClusiveUser, page: str, version_count: int):
                 if h.type.can_show(page, version_count)
                 and h.ready_to_show()]
 
+
+class CallToAction(models.Model):
```
waharnum, line 111: We'd discussed in the tech meetings that we might want to include the content for the CTA in the database itself, so that adding new CTAs doesn't require modifying template code. Should this be considered for a future enhancement?

waharnum (changes requested):
I think this should have one change (add the new fixtures to the local install scripts), but otherwise LGTM.

src/tips/models.py
bgoldowsky: Yes, I could imagine in the future that there we could add a "simple" CTA type that would work this way. The current ones all have unique markup, including buttons, images, links, etc, that would not be easy to implement as database content.

PR #115, opened 2021-07-30
bugfix/CSL-1081-theme-application

waharnum:
This should improve the theming behaviour of the Reader when first loading; it's a patch rather than a fix, as the underlying cause still seems to be an issue with the availability of custom styles when first loading and configuring the Reader. (see the diagnostic branch at https://github.com/cast-org/clusive/tree/diagnostic/CSL-1081-theme-application) 

This patch puts a second call to `applyUserSettings` behind a 2 second timeout when first loading the Reader, to ensure user settings get applied, albeit on a short delay, until the underlying Reader issue can be fixed.

This also updates the Reader version to the latest `1.9.0`.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -361,9 +361,9 @@
       }
     },
     "@d-i-t-a/reader": {
-      "version": "1.8.7",
-      "resolved": "https://registry.npmjs.org/@d-i-t-a/reader/-/reader-1.8.7.tgz",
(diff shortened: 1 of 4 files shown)


waharnum:
Update: further testing shows that Reader `1.9.0` seems to have eliminated the bug; I've removed the redundant `applyUserSettings` call on the time out, but if we continue to see this issue on QA we should double check that QA has had the dependency properly update from 1.8.7 to 1.9.0

PR #130, opened 2021-09-08
CSL-1088: Fix change-password authentication backend

klown:
Possible fix for the server error reported in [CSL-1088](https://castudl.atlassian.net/browse/CSL-1088).

I ran into a similar problem when trying to change the password of one of my test accounts in my development environment, where the stack trace indicated a missing authentication backend.  This PR fixes that by specifying Django's `django.contrib.auth.backends.ModelBackend`.


bgoldowsky:
Works for me!

PR #134, opened 2021-09-14
NOJIRA: add port in get_add_scope_redirect_uri()

klown:
@bgoldowsky I ran into this last week, and put together a quick fix.  I can open a JIRA for it if you like, but it's just a few lines of code to see if there is a port and to add that to the redirect uri as appropriate.


klown:
Closing:  the port is stored in the database in a Sites/Sites record as part of the domain name.

PR #136, opened 2021-09-20
NOJIRA: Update clusiveTTS to to improve what is considered a readable element

mbrambilla:
Current clusiveTTS misses many items in dashboard and library, probably other places too.
Examples:
  - library: the author's names are not read
  - dashbord: text in the card for the reaction graph is missed

### DIFF
--- a/src/frontend/js/tts.js
+++ b/src/frontend/js/tts.js
@@ -264,6 +264,60 @@ clusiveTTS.readQueuedElements = function() {
     }
 };
 
+clusiveTTS.wrap = function(toWrap, wrapper) {
+    'use strict';
+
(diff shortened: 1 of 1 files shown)


mbrambilla:
Also reduces jQuery use.

Still looking into improving content swap to reduce bouncing

mbrambilla:
Not perfect due to multiple browser inconsistencies around text selection and how that gets reported. Add onto that the inconsistencies of browser-based TTS, and hope for the best.

There are issues depending on how selections are made and where they start and end for the readings, but many of these already existed before, but now more portions of text are readable.

This does not fix the highlighting, rate, voice, or settings change issues currently reported.

Tested across multiple devices and browser, and seems to work well.

PR #137, opened 2021-09-21
Feature/CSL-1222 Sync to update Google classroom roster

klown:
This branch is for the Google classroom roster synchronization work.

- [x] Update "Manage" page to display roles of each individual in the period roster.
- [x] Track Google user id as part of initial roster import.
  - [x] Track Google user id during initial SSO login/register

These first three tasks above are separated out into PR #139 

- [x] Add a "Sync with Google Classroom" link/button
  - Link is present, but clicking it is circular and simply returns to the current "Manage" page
  - Will flow to "Google Roster Sync" page
- [x] Add a "Google Roster Sync" page that displays:
  - All current members (students, co-teachers) of the period
  - New members of the Google class not in the current period
  - ~~Changes to user's name~~
  - Changes to user's email
  - [x] "Save Changes" link to:
    - [x] add new members to the period, possibly creating new Clusive users,
    - [x] update email changes for existing Clusive users.
    - [x] return to the "Manage" page when done.
  - [x] "Cancel" link

JIRAs:
- Design: https://castudl.atlassian.net/browse/CSL-1221
- Implementation: https://castudl.atlassian.net/browse/CSL-1222


bgoldowsky:
Problem?
image

klown:
> This may be because my existing users don't have their google ID set in their ClusiveUser records. Is there a way we can copy that over for any existing google users?

Sounds like another task?  For a few users, use the admin interface and 

> Problem? ...

Looks like it.  What is Fnord's status of in the period roster and in the google classroom roster before clicking "Synchronize ..."?  And, does Fnord have Clusive account at that point, that is, is there is a Fnord User and ClusiveUser?

klown:
>> This may be because my existing users don't have their google ID set in their ClusiveUser records. Is there a way we can copy that over for any existing google users?

> Sounds like another task? For a few users, use the admin interface and copy the Uid from the corresponding SocialAccount. But, we'd want an automated process to update the external IDs for the long term.

There is code in `finish_login()` that updates the `external_id` for SSO users.  But, that works only if they happen to login.  If they don't but (1) are pre-external_id SSO users and (2) the roster synchronization is run, then the dependency on that `external_id` is flawed.

Could this be handled with a database migration file?  The value of the external id is present for all SSO users in their SocialAccount.  It's matter of copying it over to their ClusiveUser record.  I'll look into whether this is good way to go.  It may even be better to make a one-to-one relationship between the ClusiveUser `external_id` and the SocialAccount `uid`.

bgoldowsky:
Thanks for looking into this.

I think the migration for case 1 would be straightforward.  I've written migrations that do queries and set initial values, eg assessment migrations 11 and 12; it's not much work beyond just coding the python function to determine the right value.

For case 2, there are a couple of things that come to mind:

- I'l check and see if there are any of these on the live site that are not just test accounts. maybe we don't need to worry about it at all.
- if we do: when you run the comparison of existing roster to new roster as part of the sync process, if there are any users in the existing roster who do not have an `external_id`, then compare those users based on their email address instead (and make sure that `external_id` gets set for next time)

PR #138, opened 2021-09-28
CSL-1222 Track Google ID

klown:
This branch is part of the Google classroom roster synchronization work, but is limited to the code for tracking Google identifiers of users imported from Google classroom.  It also tracks a user's Google id when the sign up for Clusive using Google SSO.

- [x] Update "Manage" page to display roles of each individual in the period roster.
- [x] Track Google user id as part of initial Google classroom roster import.
- [x] Track Google user id as part of initial SSO login/register

JIRA: https://castudl.atlassian.net/browse/CSL-1222


klown:
Branch created from the wrong spot as it includes some of the actual sync code.
Closing and deleting.

PR #140, opened 2021-09-30
NOJRA: clusiveTTS - handle mid-read settings changes and support localization

mbrambilla:
Handles change voice or rate settings while clusiveTTS is reading.  Also includes support for looking up the DOM tree for 'lang' attributes and will try to switch to an appropriate voice if one is reported.

A little janky due to delayed nature of the speechSynthesis methods - so words can be re-read, skipped, partially announced and the re-read.

Handling is off when using a voice that does not report `onBoundary` events (such as Google Voices), but the rate and voice changes will eventually occur when the next `utterance` is created.  Word-by-word highlighting also does not work with these voices.

Questions:
- [ ] Should we only support voices reporting `localService: true`?
  - This is how R2D2BC is handling voices.
  - Android does not have default local voices (?)
  - Have not checked MacOS or iOS devices yet.

### DIFF
--- a/src/frontend/js/tts.js
+++ b/src/frontend/js/tts.js
@@ -8,7 +8,8 @@ var clusiveTTS = {
     synth: window.speechSynthesis,
     elementsToRead: [],
     region: {},
-    currentVoice: null,
+    voiceCurrent: null,
+    voiceLocal: null,
     voiceRate: 1,
     textElement: null,
     copiedElement: null,
@@ -16,6 +17,7 @@ var clusiveTTS = {
     userScrolled: false,
     readerReady: false,
     isPaused: false,
+    hasSettingUpdate: false,
     utterance: null
 };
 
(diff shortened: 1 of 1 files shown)


mbrambilla:
- [Chromium- Issue 521666: speechutterance onboundary event doesn't fire](https://bugs.chromium.org/p/chromium/issues/detail?id=521666)
- [CSL-540](https://castudl.atlassian.net/browse/CSL-540) has a table voices used in Clusive's translate feature vs what can be found in browsers.
- On those same devices- # of local voices returned from `speechSynthesis.getVoices().filter(voice => voice.localService === true);`
  - Safari/MacOS: 48
  - Safari/iOS: 38
  - Firefox/Windows: 2
  - Chrome/Windows: 3
  - Edge/Windows: 3
  - Chrome/Android: 54  (run query twice to get the list)

mbrambilla:
Windows users might need to get directed to the following for instructions on how to install more voices locally.
 https://support.microsoft.com/en-us/topic/how-to-download-text-to-speech-languages-for-windows-10-d5a6b612-b3ae-423f-afa5-4f6caf1ec5d3

PR #154, opened 2021-10-21
Feature: Bookshare OAuth2 access

klown:
OAuth2 workflow from Clusive (client) to Bookshare (provider).  This connects the currently logged in user to their Bookshare account.

Note:  there is no UI to trigger the OAuth2 workflow.  The Clusive endpoint that triggers the process is:
`/accounts/bookshare/login?process=connect`

Related JIRAs:
- [Proof of concept connecting Clusive to Bookshare](https://castudl.atlassian.net/browse/CSL-1308)
- [Plan info architecture and the experience for Bookshare search](https://castudl.atlassian.net/browse/CSL-1284) 
- [Wireframe Bookshare search](https://castudl.atlassian.net/browse/CSL-1285)


src/oauth2/bookshare/urls.py
```diff
@@ -0,0 +1,15 @@
...
+#import allauth
+
+from .provider import BookshareProvider
+
+urlpatterns = default_urlpatterns(BookshareProvider) + [
+    # Map 'socialaccount_connections' to Clusive's reader page.  This is because
+    # allauth hard codes a reverse('socialaccount_connections') to handle the
+    # end of the workflow triggered by '.../login?process=connect'.  By adding
+    # this path, the end of that workflow will end up on Clusive's reader page.
+    path('reader', ReaderIndexView.as_view(), name='socialaccount_connections')
```
bgoldowsky, line 14: This should be a redirect view rather than an extra URL for an existing page.

src/roster/signals.py
```diff
@@ -120,15 +120,14 @@ def update_social_email(sender, **kwargs):
...
         clusive_user.user.email = social_email
         clusive_user.user.save()
 
-    try:
-        email_address = EmailAddress.objects.get(user_id=clusive_user.user.id, primary=True)
-        if email_address.email.lower() != social_email:
+    # Update only the primary email address record for this clusive user
+    email_addresses = EmailAddress.objects.filter(user_id=clusive_user.user.id)
+    for email_address in email_addresses:
+        if email_address.primary and email_address.email.lower() != social_email:
```
bgoldowsky, line 131: I don't understand the reason for this change. I think the 'try' was put in to avoid errors that actually happened, and the more specific query should be more efficient.

bgoldowsky (changes requested):
I haven't tested this yet but it looks good other than a couple of comments I put inline.

src/roster/signals.py
klown: Thanks for catching this.  The reason was that the exception actually occurred  -- `get()` found two records.  The reason for that was my Bookshare provider class generated a Bookshare `EmailAddress` record with its `primary` value set to `True`. As far as the database is concerned, there is no contradiction in having multiple primary `EmailAddress`es for the same user.  The fix was (1) to change the `EmailAddress` generation to mark a Bookshare email address as not primary, and (2) to change the logic here in `signals.py` so that even if there are multiple `primary` `EmailAddress` records, then `filter()` will not throw, and only the primary ones will be updated.

But, at the moment, we only support a Bookshare connection involving a non-primary email.  The second change is overkill in that case.  I'll revert back to the former `try: get() ... except: ...` logic.

src/oauth2/bookshare/urls.py
klown: I looked into whether `allauth` supports a "next" request parameter.  It does, using the value as a redirect at the end of the authorization/access-token workflow.  Thus, you can start the workflow with, e.g., `/accounts/bookshare/login?process=connect&next=/dashboard`, and the user will end up on the Dashboard page after successfully connecting to their Bookshare account.  That's actually quite useful: Depending on which page the user initiates authorization, the url can specify an appropriate end point.

However, given the `process=connect` parameter, `allauth` insists on running `reverse('socialaccount_connections')` as part of its connections workflow.  Its `DefaultSocialAccountAdapter` class, specifically its `get_connect_redirect_url()` method, is where happens.  I can find no way to change that behaviour by a simple change of setting or view, other that to map `socialaccount_connections` to a known Clusive url.

The `allauth` [documentation](https://django-allauth.readthedocs.io/en/latest/advanced.html#custom-redirects) says to:
1. subclass `DefaultSocialAccountAdapter` and override its `get_connect_redirect_url()' method
2. set `SOCIALACCOUNT_ADAPTER` to the new class in the project's `settings.py`

Note that 2 is "global" in the sense that the subclass is used for all `allauth` workflows.  In Clusive's case, that means it's also used for Google SSO.  Practically, this is not a problem since Google SSO dos not involve `process=connect`.  But, it is something to keep in mind for other authorization scenarios we might implement in the future.

Another route is to use `allauth`'s default setup for the `socialaccount_connections` endpoint, but that entails incorporating its associated social connections html template, and everything that it imports.  I haven't explored this solution much since I assumed Clusive would not want that `allauth` page to be available to users.

The latest commit uses the subclass/override technique.

klown: I had another thought:  Instead of implementing the sublcass, etc., change this particular url pattern to:

> `path('bookshare_connected', views.bookshare_connected, name='socialaccount_connections')`,

where `bookshare_connected()` has a default redirect to somewhere suitable.  But, the presence of a `next=/...`parameter in the original request would have precedence.

Maybe that's what you meant in the first place?

bgoldowsky (approved):

klown:
Closing since this has been merged into the main Bookshare development branch (`feature/CSL-1412-bookshare-search`).

PR #159, opened 2021-11-04
CSL-1370: Force a default TTS voice

mbrambilla:
New implementation:

-  Assume base language of `en`
-  Check user/browser languages for region specific `en-*` variants - default to first voice on match based on order of preference - as reported by `navigator.languages`
- otherwise default to first `en-US` voice
- default to first `en` voice if no `en-US` voice available
-  Remove the 'Choose…' option from list since we are forcing a voice

I think it should cover most of our current/intended user base - there could be some issues for international users - but most language related-items are already hard-coded to an `en` base.  Testing on my environments seems to be working well otherwise.

### DIFF
--- a/src/frontend/js/tts.js
+++ b/src/frontend/js/tts.js
@@ -782,6 +782,7 @@ clusiveTTS.getVoicesForLanguage = function(language) {
     'use strict';
 
     var voices = [];
+    var defaultVoices = [];
     if (!language) {
         return voices;
     }
@@ -790,12 +791,16 @@ clusiveTTS.getVoicesForLanguage = function(language) {
         var voiceLang = voice.lang.replace('_', '-').substring(0, language.length);
         if (voiceLang === language) {
             if (voice.default) {
-                voices.unshift(voice); // Put system default voice at the beginning of the list
+                defaultVoices.push(voice);
             } else {
                 voices.push(voice);
             }
         }
     });
+
+    // Put system default voice at the beginning of the list
+    voices = defaultVoices.concat(voices);
+
     return voices;
 };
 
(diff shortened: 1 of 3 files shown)


mbrambilla:
Android is still an edge case at the moment since it lists/reports multiple available voices, but only uses the voice selected in OS settings.  Still looking to see if there is a way to determine what that is, but coming up empty. 

You would think the voice would be reported with `SpeechSynthesisVoice.default: true`, but this is not the case.

src/shared/templates/shared/partial/modal_settings.html
```diff
@@ -200,7 +200,10 @@ <h2 class="modal-title h5">
                                 <fieldset>
                                     <legend class="setting-title">Voice</legend>
                                     <select class="cislc-modalSettings-readVoice form-control">
+                                        {% comment %}
```
bgoldowsky, line 203: Can take out this commented region now.

bgoldowsky:
One small logic error. Otherwise looks good to go.

src/frontend/js/tts.js
```diff
@@ -830,6 +835,46 @@ clusiveTTS.setCurrentVoice = function(name) {
...
+
+    // Reference: https://stackoverflow.com/questions/1043339/
+    // Possibly useful in future? - https://stackoverflow.com/a/29106129
+    // eslint-disable-next-line compat/compat
+    var userLanguages = window.navigator.languages || [window.navigator.language || window.navigator.userLanguage];
+    var defaultVoice = null;
+    var langVoices = [];
+
+    userLanguages.some(function(lang) {
+        if (lang.substring(0, 1) === 'en' && lang.length > 3) {
```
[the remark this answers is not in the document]
mbrambilla: Oh that could have been bad.  Revised with recommended `lang.startsWith('en-')`

src/shared/templates/shared/partial/modal_settings.html
mbrambilla: Removed

didactica:
Hi, can you guys help us?, we need to chage the dictionary to spanish, do you do that  in Clusive?

PR #163, opened 2021-11-25
Add reduce motion preference

klown:
This adds:
- a preferences panel to the preferences side bar containing a "Reduce motion" switch,
- a database two way binding to store/update the preference and switch state,
- a default preference of "Off",
- an enactor binding that responds to preference changes.

JIRA: https://castudl.atlassian.net/browse/CSL-1409


mbrambilla:
Related updates with CSS and JS were already merged into `development` branch.  Reference merge commit: 490a5c752ac74e2d050205a8923c5918dc673343

The JS calls for the Clusive specific reduced-motion:
-  Enable reduce-motion: `cisl.prefs.setReducedMotion()`
- Resume full animations: `cisl.prefs.unsetReducedMotion()`
 
Reference: https://github.com/cast-org/clusive/blob/development/src/frontend/js/clusive-prefs-panel.js#L257-L272

The Clusive specific reduced-motion setting will currently only affect the blinking mascot image at the top of the Dashboard page, and the confetti-cannon that fires when selecting the final word-rating in the dialog that shows on first read of a publication or a star rating is completed..  Reduced motion result will be a static image instead of the animated PNG, and no confetti being shown where appropriate.

klown:
thanks @mbrambilla; this was helpful.

I copied the logic of the `setReduceMotion()` and` unsetReduceMotion()` functions into the fluid preference enactor function, and then removed them since I did not see that they were used anywhere, and was concerned about possible synchronization issues.

I also put in a function to make a @media check and to consult the OS level "reduce motion" preference on login as part of the logic that restores a user's preference from the database.

I'm currently looking into how to hook in changes to the OS level "reduce motion" preference so that the preferences panel and its associated preference is updated as needed.

klown:
@bgoldowsky Based on your [JIRA comment](https://castudl.atlassian.net/browse/CSL-1409?focusedCommentId=31365) about keeping the clusive and OS level preferences independent, I removed the function that examined the OS level preference and, if on, switched the clusive preference to be in sync.  With that removal the two preferences are independent.

Ready for review.

klown:
I found an infusion built-in enactor, `fluid.prefs.enactor.styleElements`, that does most of the work to realize style changes and it seemd appropriate for the "reduce motion" clusive preference.  I've modified the code to make use of it.

PR #168, opened 2021-12-09
Add reduce motion preference

klown:
@bgoldowsky @mbrambilla Here are the changes to the former "Reduce motion" switch, inverting it to "Animations (on/off)"

JIRA decision: https://castudl.atlassian.net/browse/CSL-1409?focusedCommentId=31635 and https://castudl.atlassian.net/browse/CSL-1409?focusedCommentId=31636

### DIFF
--- a/src/frontend/js/clusive-prefs-modal-settings.js
+++ b/src/frontend/js/clusive-prefs-modal-settings.js
@@ -96,9 +96,9 @@
                 args: ['{change}.value', 'preferences.cisl_prefs_glossary', '{that}'],
                 excludeSource: 'init'
             },
-            'modalSettings.reduceMotion': {
+            'modalSettings.animations': {
                 funcName: 'cisl.prefs.modalSettings.applyModalSettingToPreference',
-                args: ['{change}.value', 'preferences.cisl_prefs_reduceMotion', '{that}'],
+                args: ['{change}.value', 'preferences.cisl_prefs_animations', '{that}'],
                 excludeSource: 'init'
             },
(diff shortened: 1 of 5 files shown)


bgoldowsky:
@klown the switch in the UI needs to be inverted as well.  Not 
`On <--> Off`
but
`Off<-->On`,
highlighting in red when "On".

klown:
@bgoldowsky the latest changes address your comment.  And, I've made a number of other modifications based on our decision in slack to change the name of the preference to match the UI.

One other issue came up.  The infusion component that changes the CSS, `styleElements` applies a style when the UI switch is on, and removes it when it's off -- the opposite of what it should do.  I added an override of its `handleStyle()` invoker to flip the use of its `applyStyle()` and `resetStyle()` functions.  But, there may be a simpler way to configure `styleElements` so I will confirm with the infusion folks.

The other way to avoid that override is to change the sense of the CSS class used and when it is added/removed, namely `.clusive-reduced-motion`.  That might be a bigger can of worms; @mbrambilla  would know better.

klown:
@bgoldowsky @mbrambilla I wrote:

>  ... there may be a simpler way to configure styleElements so I will confirm with the infusion folks.

I had a chat with @jobara about overriding `handleStyle()` and he thought it was a good approach.  This is done from my end.

bgoldowsky:
Looks good to me. I think having the CSS element remain as reduced-motion makes sense too - consistent with how the OS and CSS treat this, though different from our UI.  Will merge.

PR #169, opened 2021-12-10
Word bank: Add drag and drop

mbrambilla:
Click and hold the word for slightly less than 1/2 a second to trigger the drag handling.  A solid border will surround the item once enabled.

Drag is limited to the X (horizontal) axis - so no reordering and words are shown alphabetically listed anyway.

Hold delay used so that the word, movement, and remove items can still be clicked.

Updated the confetti to fire from the word (instead of the arrow) after the slide animation has completed.  Dropping a word disables the slide animation and fires the confetti if the drop is into a new column (different from the originating one).


mbrambilla:
Preview available: https://preview.cast.org/clusive/html/wordbank.html

PR #175, opened 2022-01-11
NOJIRA: POC to move focus from TOC into reader frame

mbrambilla:
Accessibility feedback showed that screen reader does not move to the location within the text, and there is no feedback that something occurred, such as the content within the reader frame has updated and scrolled into view.

Testing with NVDA/Windows and VoiceOver/MacOS. 

- Both VoiceOver and NVDA simulate clicks when activating a link from within the TOC.  No good way of telling a simulated click apart from mouse/touch click so this would need to apply to all users, not just keyboard or screen reader users.
- At least one layer of setTimeout seems to be needed to move focus into the frame.  Current testing has two layers due to bouncing around ideas.
- `tabindex=-1` seems needed to get focus properly moved for the screen readers.
- NVDA can still get hung up reading portions of the Clusive wrapper, sometimes just the TOC button, before reading focus into the frame..

Not necessarily happy with how this works, but this shows it is somewhat possible.  This method seems 'dirty', and might need additional improvements made upstream.

### DIFF
--- a/src/frontend/js/toc.js
+++ b/src/frontend/js/toc.js
@@ -13,6 +13,8 @@ var TOC_EMPTY = '#tocEmpty';
 var NOTES_TAB = '#notesTab';  // links to #notesPanel
 var NOTES_CONTAINER = '#notesList';
 
+var TOC_focusSelector = null;
+
 // eslint-disable-next-line no-unused-vars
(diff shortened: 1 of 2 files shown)


mbrambilla:
Replaced by #226

PR #190, opened 2022-02-08
Feature/CSL-1460 Implement customized vocabulary words for readings

klown:
Implement customized vocabulary words -- branch/PR for all [CSL-1460](https://castudl.atlassian.net/browse/CSL-1460) related tasks
- [x] Create database structure for vocabulary words ([CSL-1547](https://castudl.atlassian.net/browse/CSL-1547)),
- [ ] In Progress: Add vocab word editing to customizations page ([CSL-1548](https://castudl.atlassian.net/browse/CSL-1548)),
- [ ] Make use of custom vocab words in choosing "Just right challenge" words and cued words ([CSL-1549](https://castudl.atlassian.net/browse/CSL-1549)).


klown:
Added implementation for the second task, "Add vocab word editing to customizations page" ([CSL-1548](https://castudl.atlassian.net/browse/CSL-1548)).

src/library/templates/library/edit_customization.html
```diff
@@ -103,5 +126,146 @@ <h2>Edit {{ customization.title }} {% if is_new == 'true' %}(New){% endif %}</h2
...
+    });
+
+    $('.delete-word-button').click(function (e) {
+        deleteVocabularyWord(e);
+    });
+
+    function deleteVocabularyWord (e) {
+        e.preventDefault();
+        var parentDiv = $(e.currentTarget).parent();
+        var wordToDelete= $(e.currentTarget).attr('id').replace(/^delete_/,'').replace('_', ' ');
```
bgoldowsky, line 148: You can avoid having to do string substitutions like this by using a data attribute, eg data-word="{{word}}",  wordToDelete=$(e.currentTarget).data('word')

src/library/templates/library/edit_customization.html
```diff
@@ -103,5 +126,146 @@ <h2>Edit {{ customization.title }} {% if is_new == 'true' %}(New){% endif %}</h2
...
+            var condition = addCondition(newWord);
+            if (condition === 'addNew' || condition === 'addAndRemoveFromDeleted') {
+                var delete_button_id = 'delete_' + newWord.replace(' ', '_');
+                // Add the new <div> for the "new" vocabulary word.
+                $('.vocabulary-word-list').append(`
+                    <div class="col-lg-4 vocabulary-word new-vocabulary-word">
+                        ${newWord}  <button id="${delete_button_id}" class="link-btn">x</button>
+                    </div>
+                `);
+                // Hook up the delete listener for the new word's delete button.
```
bgoldowsky, line 191: Using a [delegated handler](https://learn.jquery.com/events/event-delegation/) would avoid the need for this.

src/library/templates/library/edit_customization.html
```diff
@@ -103,5 +126,146 @@ <h2>Edit {{ customization.title }} {% if is_new == 'true' %}(New){% endif %}</h2
        $input.val($input.val() + text);
        $input.focus();
     });
+
```
bgoldowsky, line 129: The JS portion of this should eventually be moved into one of the JS files.

bgoldowsky (commented):
No need to make a lot of changes if this is all working, but it could probably be simplified by maintaining the data as javascript arrays, and then just writing it into the pipe-separated lists in the input fields when the user hits submit.  Then there would be a lot less parsing and unparsing.  The current words could be passed as a list directly in the template.

klown:
> No need to make a lot of changes if this is all working, but it could probably be simplified by maintaining the data as javascript arrays, and then just writing it into the pipe-separated lists in the input fields when the user hits submit. Then there would be a lot less parsing and unparsing. The current words could be passed as a list directly in the template.

After I finished this first go, I starting thinking along the same lines, namely, on loading the page, create arrays for each list; then do all the manipulations on those arrays; and, on submit, transfer the arrays back to the `input`s and then allow the submit to continue.  I started to look into making those changes, but decided to wait until review.

If there's time, and you think it's good idea, I could re-write it that way.  I think it would be easier to maintain going forward.

src/library/templates/library/edit_customization.html
klown: Great idea!  I've used the `@data` attribute in the past, but it didn't occur to me this time, nor for this purpose.  But it's definitely cleaner.

src/library/templates/library/edit_customization.html
klown: A previous version had the handler higher up and relied on bubbling, but, whenever a bug occurred the click would bubble such that the form would submit and navigate away from the edit customization page -- there was little information left around to indicate what went wrong.  To counter that, the handler stopped propagation right at the delete button, and that entailed having a handler for each delete button.

But now that things are more stable, yes, a single handler is better, and put it higher up in chain.

src/library/templates/library/edit_customization.html
klown: Right.  Probably `frontend.js`.  I don't see any JS file specifically for the library pages.  There are some library related scripts in both of `frontend.js` and `toc.js`.

PR #199, opened 2022-02-24
CSL-1607: Detect and display organizational Bookshare OAuth2 accounts

klown:
As part of the Bookshare connect workflow, obtain the `studentStatus` field from Bookshare and store it in the `extra_data` of the user's `SocialAccount`.  Thereafter, that status can be queried to determine if the linked Bookshare account is a single user account or an organizational account; and, display its organizational status on the user's account profile page, as well as the name of the organization.

JIRA: https://castudl.atlassian.net/browse/CSL-1607


src/oauth2/bookshare/views.py
```diff
@@ -105,5 +117,10 @@ def bookshare_connected(request, *args, **kwargs):
     request.session['bookshare_connected'] = True
     return HttpResponseRedirect(reverse('reader_index'))
 
+def is_organizational_account(request):
+    bookshare_adapter = BookshareOAuth2Adapter(request)
+    return bookshare_adaptor.is_organizational_account()
```
bgoldowsky, line 122: variable spelled differently than line above

klown:
@bgoldowsky The latest code is demonstrates the two techniques for determining the type of account.  It currently uses the "retrieve membership list" technique when Clusive connects to the user's Bookshare account.  The unit tests show how the user preferences can be used instead. And, it is relatively easy to switch between the two where [the code makes the connection](https://github.com/cast-org/clusive/blob/15e9f5f66204c0c27915882cb9c59fd4c314ce7e/src/oauth2/bookshare/views.py#L78).  Instead of calling `get_organization_members()`, call `get_user_type()` and check the return value for `INDIVIDUAL` vs. `ORGANIZATIONAL`.

A note about the unit tests:  they require access keys which cannot be published to this repository.  Instead, the tests look for and input the keys from a file name `keysForTesting.json`.  Each developer will have to create this file.  There is a lengthy [comment](https://github.com/cast-org/clusive/blob/15e9f5f66204c0c27915882cb9c59fd4c314ce7e/src/oauth2/bookshare/tests.py#L81) in the test code explaining this.

PR #201, opened 2022-02-26
Feature/csl 1401 starred reading

lmccormack:
Updated code to include the star rating - this needs another set of eyes.

Event logging is not working properly - may not be called and tags docs may be out of date (or not found properly - wiki is old and nothing in jira).

Filtering for starred rating has not been done and is slightly out of scope.


requirements.txt
```diff
@@ -1,5 +1,5 @@
 dawn==0.11.0
-django==2.2.24
+django~=2.2.13
```
bgoldowsky, line 2: Not sure why this looks like it's getting downgraded

requirements.txt
```diff
@@ -20,10 +20,15 @@ pandoc==2.0a4
...
 whitenoise==4.1.3
 wordfreq==2.3.2
 -e git+https://github.com/IMSGlobal/caliper-python@1.1.0.2#egg=imsglobal_caliper
 -e git+https://github.com/ouhouhsami/django-progressbarupload#egg=django-progressbarupload
 -e git+https://github.com/cast-org/django-session-timeout#egg=django-session-timeout
+
+requests~=2.25.1
+aniso8601~=9.0.1
+rfc3986~=1.4.0
+setuptools~=40.8.0
```
bgoldowsky, line 34: I don't understand how any of the changes in requirements.txt are related to this PR

src/frontend/js/frontend.js
```diff
@@ -766,12 +766,34 @@ function starredButtons() {
...
+    return $.post('/library/setstarred', {
+        book: book,
+        starred: isActive
+    })
+        .fail(function(err) {
+            console.error('Set location API failure!', err);
+        });
+}
+
+
```
bgoldowsky, line 796: Is this used somewhere?

src/library/models.py
```diff
@@ -461,6 +471,16 @@ def latest_for_user(cls, user: ClusiveUser, max=10):
...
+    # @classmethod
+    # def get_starred_values(cls, user: ClusiveUser):
+    #     """Return a QuerySet for Paradatas for a user """
+    #     return Paradata.objects.filter(user=user)
+    #
+    # @classmethod
+    # def get_starred_values_for_book(cls, user: ClusiveUser, book: Book):
+    #     """Return a QuerySet for Paradatas for a user """
+    #     return Paradata.objects.filter(user=user, book=book)
+
```
bgoldowsky, line 483: Unneeded?

src/library/templates/library/partial/book_starred.html
```diff
@@ -0,0 +1,43 @@
...
+Library list
+<div class="card-library-favorite">
+    <div class="btn-check" data-cfw="tooltip" data-cfw-tooltip-title="Not starred" data-cfw-tooltip-original-title="" title="" id="cfw-tooltip-429959">
+        <input id="starred0" type="checkbox" class="btn-check-input btn-check-starred" aria-label="Starred">
+        <label for="starred0" class="btn btn-icon btn-starred link-above">
+            <span class="icon-starred" aria-hidden="true"></span>
+        </label>
+     </div>
+</div>-->
+<!--Library grid
```
bgoldowsky, line 26: All this commented stuff will actually be rendered to the browser. It should be deleted before we merge this.

src/library/templates/library/partial/book_starred.html
```diff
@@ -0,0 +1,43 @@
+{% load static library %}
+
+{% if starred|lower == 'true' %}
```
bgoldowsky, line 3: You are passing a boolean so this can just be {% if starred %}

src/library/templates/library/partial/library_card.html
```diff
@@ -1,6 +1,6 @@
 {% load static library %}
 <div class="card card-library">
-    <div class="card-header">
+    <div class="card-header has-favorite" >
```
bgoldowsky, line 3: Class attribute should presumably only be included if show_starred is true.

src/pages/templatetags/library.py
```diff
@@ -42,3 +42,7 @@ def highlight(text, search):
 @register.filter(is_safe=True)
 def period_names(assignment_list):
     return ", ".join([p.period.name for p in assignment_list])
+
+@register.filter(is_safe=True)
+def starred_val(paradata_set):
+    return ", ".join([str(p.starred).lower() for p in paradata_set])
```
bgoldowsky, line 48: I think we can avoid the need for this unbeautiful thing (though learning how to create filters is valuable). Let me play around with it a bit so that I don't suggest something that doesn't work.

bgoldowsky (changes requested):
I haven't looked at the event logging part yet, but inserted some comments to help clean up the code.

bgoldowsky:
Here's an implementation that does not require the starred_val filter, uses standard django template functionality instead:
```
{% with  book.paradata_list|first as paradata %}
   {% include "library/partial/book_starred.html" with starred=paradata.starred book_id=book.id %}
{% endwith %}
```

src/library/templates/library/partial/library_card.html
lmccormack: Both library card mockups have the has-favorite class whether the current state is checked or not.

requirements.txt
lmccormack: requirements.txt now the same as the development branch

src/library/models.py
lmccormack: removed commented code

src/library/templates/library/partial/book_starred.html
lmccormack: removed commented code

src/library/templates/library/partial/library_card.html
mbrambilla: `.has-favorite` is needed to give the title a margin on the right side to keep the text from flowing under the star button.  Needs to be there only if the star button is present.  Not using the check here adds the unneeded margin on the dashboard cards.

src/frontend/js/frontend.js
lmccormack: removed this

src/library/templates/library/partial/book_starred.html
lmccormack: Modified to improve if using boolean

src/library/templates/library/partial/library_card.html
lmccormack: I missed this the first time through - now class has-favorite is not used when the starred value is not present (e.g., dashboard).

PR #208, opened 2022-03-11
CS-1646 Add starred filter for library books

lmccormack:
Questions: 

Starred books are not immediately taken out of the Starred List - view is not refreshed (could change mind and the book is gone).  Is this the correct behavior?

What is the order of the filter?  Order is: period, starred, public, my, all.  Is this ok?

Similar to my books, starred is a filter even when there are no starred readings.  Is this ok?


bgoldowsky:
Looks perfect to me.

PR #214, opened 2022-03-18
CSL-1646: Add migration for new STARRED LibraryView

klown:
@boris, @lmccormack I happened to notice that my local copy of the `development` branch is missing a migration for the changes to the `ClusiveUser` model for starred readings.  After a quick search here on github, I couldn't find the migration in the `development`  branch, so I made one.

If it is actually missing, this PR fixes that; otherwise, you can ignore it.


lmccormack:
The migration file [0048_paradata_starred](https://github.com/cast-org/clusive/blob/feature/CSL-1401_starred_reading/src/library/migrations/0048_paradata_starred.py).py  was checked in as part of CSL-1401 20 days ago.  Did you get the this migration?

klown:
> The migration file [0048_paradata_starred](https://github.com/cast-org/clusive/blob/feature/CSL-1401_starred_reading/src/library/migrations/0048_paradata_starred.py).py was checked in as part of CSL-1401 20 days ago. Did you get the this migration?

Thanks @lmccormack  -- I have that one.  But, that's not the migration that this is.  This is a change to the `ClusiveUser`, where it adds the "Starred readings" choice to the library view popup.  The one you reference adds a boolean to the `Paradata`.

What started this:  I was changing the ClusiveUser model for Bookshare organizational accounts and when I made the migration for my changes, the one for the library view popup showed up in my migration file.  The problem  is, I did not make any changes to the library view aspect of the ClusiveUser model.  That ultimately led to this PR.

lmccormack:
It looks like missing this migration happened before as noted in the check in for [0040_auto_20210925_0141.py](https://github.com/cast-org/clusive/blob/development/src/roster/migrations/0040_auto_20210925_0141.py).  If I run the makemigrations, I get the same file.  I'm pulling in your request.  Thanks!

PR #220, opened 2022-04-14
CSL-1662 time tracking with beacon api

klown:
JIRA: https://castudl.atlassian.net/browse/CSL-1662

Uses the browser's `navigator.sendBeacon()` to record Clusive `PageHide` events.

The call to `sendBeacon()` is invoked for both document `visibilityChange` and window `pageHide` events.  An example of the latter is turning pages in the reader.  Turning pages is missed if relying solely on `visiblityChange`.  In addition, `visibilityChange` events occur only when logging out, and is missed by the `pageHide` event.

There are some outstanding issues, however.  First, on logout, the last group of `sendBeacon()` requests result in error responses.  For all but the last request, Django issues a `SuspiciousOperation` exception.  The final response is a one line "no session" error.  Nonetheless, on the plus side, Clusive's logs show the events are processed and examination of the databse shows that they were stored.

For completeness here is an example of one of the `SuspiciousOperation` stack traces.

> SuspiciousOperation at /messagequeue/
> 
> The request's session was deleted before the request completed. The user may have logged out in a concurrent request, for example.
> 
> Request Method: 	POST
> Request URL: 	http://localhost:8000/messagequeue/
> Django Version: 	2.2.24
> Exception Type: 	SuspiciousOperation
> Exception Value: 	
> ...

My guess is that the request is processed and the events are persisted, but, when Django attempts to send the response, the session has already been cleared due the simultaneous logout request and Django cannot associate the response with any session; and complains.  It appears to be harmless though.  We could look into delaying the logout response until all `messagequeue` requests have been processed.

Secondly, there is an issue of how the change to using `sendBeacon()` interacts with the `clusive.djangoMessageQueue` component's `localStorage`.

Prior to using `sendBeacon()`, Clusive `PageHide` events were synchronized with `localStorage` to make sure that the end time was recorded only once.  At the same time, the `clusive.djangoMessageQueue`'s queue is synchronized with the contents of `localStorage`.

The result was that PageHide events were stored in `localStorage`, eventually merged with the `clusive.djangoMessageQueue`, and ultimately POSTed to Clusive itself.  Switching to `sendBeacon()` means that the `PageEnd` events are not placed into `localStorage` nor the `clusive.djangoMessageQueue`.  They are now completely independent of that queue.  That might mean that the end time is sent to Clusive more than once.  However, there is only ever one such event recorded in Clusive's database.  I suspect there is only one because the `eventId` in the body of the request is used as the `id` of the event record in the database, and any update of the database writes over any existing record with that `eventId/id`.

There are two ways forward.  One is to look into using `sendBeacon()` for all of Clusive's events and discontinue `clusive.djangoMessageQueue` (while pirating some of its logic in terms of packaging the events for the requests). Another is to create a "local storage" specifically for `PageHide` events and synchronize them separately.  The former is a longer term solution.


klown:
Noting a [stackoverflow](https://stackoverflow.com/questions/59443915/django-navigator-sendbeacon-csrf-protection) article on how to satisfy django in terms of the csrf token.  The answer is to send the `data` as `FormData`, setting the form parameter `csrfmiddlewaretoken` to the value of the csrf token.  The next question is what are the implications to Clusive's `MessageQueueView.post()` in terms of handling `FormData` instead of a JSON string.

src/frontend/js/page-timing.js
```diff
@@ -54,13 +55,33 @@ PageTiming.trackPage = function(eventId) {
...
-    // Set up so that end time will be recorded
-    $(window).on('pagehide', function() { PageTiming.reportEndTime(); });
+    // Set up so that end time will be recorded when user turns pages.
+    // https://developer.mozilla.org/en-US/docs/Web/API/Window/pagehide_event
+    $(window).on('pagehide', PageTiming.windowEventListener);
+
+    // Record end time when document is hidden.
+    // https://developer.mozilla.org/en-US/docs/Web/API/Navigator/sendBeacon
+    // https://developer.mozilla.org/en-US/docs/Web/API/Document/visibilitychange_event.
+    document.addEventListener('visibilityChange', PageTiming.docEventListener);
```
bgoldowsky, line 65: I am concerned about this - visibilityChange triggers on things like just switching to a different tab in the browser, which does not mark the end of a pageview.

src/frontend/js/page-timing.js
```diff
@@ -129,27 +150,90 @@ PageTiming.recordInactiveTime = function() {
...
+        // Add the PageEnd event to the messagequeue.
+        clusiveEvents.messageQueue.add(PageTiming.createEndTimeMessage());
+        console.debug(`Queued ${PageTiming.eventId} via clusiveEvents.messageQueue.`);
+    } else {
+        console.warn('Pagehide event received, but end time was already recorded');
+    }
+};
+
+// Utility for retrieving the list of PageEnd times from window.localStorage
+PageTiming.getEndTimesProcessed = function () {
```
bgoldowsky, line 226: If we're sending each end time to the server immediately at the moment the page view ends, I do not understand why keeping a list in LocalStorage is necessary any more. That was just a way to keep the info around so that it could be sent later.

bgoldowsky (commented):
There are a couple of issues I am unclear on, can we talk this through?

src/frontend/js/page-timing.js
klown: My experience was that `visibilityChange` events only happened when you closed the "Clusive" tab or logged out.  Switching to a different tab caused `pagehide` events only, which the previous version was already tracking. 

However, let me double-check that.

src/frontend/js/page-timing.js
klown: I thought that another reason for using `localStorage` was to avoid a case where `PageTiming.reportEndTime()` was triggered more than once with the same `eventId`.  There was/is a check for that at the top of `PageTiming.reportEndTime()`.  I kept that check, but with changes to its implementation.  Here is how it used to be done:
> `if (!localStorage['pageEndTime.' + PageTiming.eventId]) { ...`

Using `sendBeacon()` does not necessarily mean the message is sent immediately.  The request is put on some internal browser queue, and the browser will send it in a timely manner (according to the documentation).  Still, since there is no way for the `PageTiming` code to tell if it was successfully sent, then maybe it doesn't actually matter if the same `eventId` is sent multiple times.  But also, empirically speaking, I have not seen the message from the `else` clause in the browser console: 
> `console.warn('Pagehide event received, but end time was already recorded');`

That implies that `PageTiming.reportEndTime()` is only ever called once per `eventId`.  I will happily remove that tracking code and its use of `localStorage`.

src/frontend/js/page-timing.js
klown: > However, let me double-check that.

Here's what I see in the log:
- `pageHide` events when switching between Clusive tabs (e.g., from "Dashboard" to "Library"),
- `PageTiming.handleFocusChange()` when switching among browser tabs (not `pageHide` as I thought above, but the machinery for tracking inactive time),
- `visibilityChange` events with`document.visibilityState = 'hidden'` when closing the "Clusive" browser tab, or on logout.  Actually, I can't see that in the log when closing the tab since the log window closes with it.  I'm making the assumption that because the `PageEvent` is updated in the database that the event occurred.

With respect to `visibilityChange`, what I see in the logs contradicts the [MDN documentation](https://developer.mozilla.org/en-US/docs/Web/API/Document/visibilitychange_event#usage_notes) stating that the event occurs when "...a user navigates to a new page, switches tabs, ...", as you noted in your comment.

On the other hand, `visibilityChange` is also [described elsewhere](https://developer.mozilla.org/en-US/docs/Web/API/Navigator/sendBeacon#sending_analytics_at_the_end_of_a_session) as the most reliable way to detect the end of session for sending diagnostics, which is what we want to use it for.

I'll see if there is a way to distinguish between moving among browser tabs vs. closing the tab entirely.

bgoldowsky: BTW there is a 'persist logs' option in the settings for the javascript console.

src/frontend/js/page-timing.js
bgoldowsky: Duplication of end time reporting is also checked for on the server side, so removing this check seems fine.

src/frontend/js/page-timing.js
bgoldowsky: Oh, but persist logs does not help if you close the tab completely. Right.

bgoldowsky: OK, in my experimentation a 'visibilityChange' event, as written on line 65, never fires.  If I listen for 'visibilitychange', all lowercase, it fires when you switch tabs, completely hide the web browser, etc.  Let's stick with pagehide for now; although I do see the warnings that say it's unreliable on mobile.

bgoldowsky (commented):
Let's get the sendBeacon change in place and defer tinkering with pagehide vs. other events for later.

klown:
Ugh!  Hoisted by a camel-case typo.

Okay, I'll remove all the code for listening for and handling `visibilityChange`, and, by implication, `visibilitychange`.

PR #222, opened 2022-04-20
NOJIRA: Update npm dependencies

mbrambilla:
Mostly grunt and front-end related dependencies to start with.
- eslint
- stylelint
- sass compilation: node-sass, postcss, etc.
- Figuration
- fixed some lint errors

Updated to R2D2BC latest release, at time of commit.

Updated `.browserslistrc` to change support to only modern browsers, dropping IE and other outdated browsers.

Fixed markup issue with dashboard page.  
Minor sidebar/content gutter spacing tweak.
Replaced custom loader with Figuration's Loader component.

Still need to check Webpack.
Cypress test appear to need some fixes.
Infusion will be not be included at this time as beacon testing is underway.


mbrambilla:
**From Webpack commit:**
- Removed expose-loader as it does not appear to be used anywhere.
- Removed websocket-extensions as it was manually added for a security issue, nested dependency has been brought up to date, so no longer needed


**From Cypress commit:**
Adds a minor work-around to get Cypress preference tests to pass, due to issue with Cypress and `<iframe>` use.

`window.parent` gets rewritten by Cypress - setting `"modifyObstructiveCode": false` in `cypress.json` is usually the fix, but doing so is causing other errors.  Went with the typeof checks for a quick fix.
Refs:
- https://github.com/cypress-io/cypress/issues/5538
- https://on.cypress.io/configuration#modifyObstructiveCode

This could use a revisit at a later time.

PR #226, opened 2022-04-25
CSL-1515: TOC and Highlight items move focus to reader when clicked

mbrambilla:
Need updated Figuration v4.3.2 in order to remove forced reset of focus to TOC button when modal is closed.  TOC/Highlight modal has been change to use 'manual' mode.

Depends on #222
Replaces #175


mbrambilla:
Ready for review.

Note that NVDA tries to be smart when `role="dialog" items close/hide.  NVDA forces focus back to the TOC button upon close of modal, even though Figuration is not doing so, but the focus does get moved properly into the reader frame.

VoiceOver focus seems to move directly into the reader frame without a brief layover on the TOC button.

mbrambilla:
Ugh, rebase gone wrong.  Closing, will create new PR.

PR #232, opened 2022-04-27
CSL-1515: TOC and Highlight items move focus to reader when clicked

mbrambilla:
Replaces #226

Note that NVDA tries to be smart when `role="dialog" items close/hide. NVDA forces focus back to the TOC button upon close of modal, even though Figuration is not doing so, but the focus does get moved properly into the reader frame.

VoiceOver focus seems to move directly into the reader frame without a brief layover on the TOC button.

### DIFF
--- a/src/frontend/js/toc.js
+++ b/src/frontend/js/toc.js
@@ -138,28 +138,31 @@ function buildTableOfContents() {
 
     if (typeof d2reader === 'object') {
         var items = d2reader.tableOfContents;
-        var has_structure = items.length > 1;
+        var has_structure = items.length > 0;
         if (items.length === 1 && items[0].children && items[0].children.length > 1) {
             has_structure = true;
         }
-        if (has_structure) {
-            $(TOC_EMPTY).hide();
-            var out = buildTocLevel(items, 0, 'toc');
-            $(TOC_CONTAINER).html(out).CFW_Init();
-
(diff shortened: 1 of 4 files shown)


src/frontend/js/toc.js
```diff
@@ -460,26 +460,121 @@ function setUpNotes() {
...
 }
 
-//  Initial setup
+// Accessibility addition to close TOC and move focus to selected item within the reader frame.
+// Uses click event because screen readers since they use simulated click events,
+// even if keyboard navigation is used.
+function tocClickHandler() {
+    'use strict';
+
+    $(TOC_CONTAINER).on('click', 'a:not([data-cfw="collapse"])', function(event) {
```
bgoldowsky, line 469: There's another click handler for - I think - the same set of links above at line 159. Would it make sense to combine them so the code is that much more clear?

src/frontend/js/toc.js
```diff
@@ -136,28 +138,31 @@ function buildTableOfContents() {
 
     if (typeof d2reader === 'object') {
         var items = d2reader.tableOfContents;
-        var has_structure = items.length > 1;
+        var has_structure = items.length > 0;
```
bgoldowsky, line 141: With this change the three lines below are redundant - has_structure will already be true.

PR #242, opened 2022-05-06
CSL-1533 Prevent negative active time from getting recorded

klown:
- "Handle" negative `active_duration` in PageTiming by setting it to zero. 
- Detect and use `PerformanceNavigationTiming` API for getting load times if available.

JIRA: https://castudl.atlassian.net/browse/CSL-1553

### DIFF
--- a/src/frontend/js/.eslintrc.json
+++ b/src/frontend/js/.eslintrc.json
@@ -2,6 +2,7 @@
   "root": true,
   "env": {
     "browser": true,
+    "es6": true,
     "jquery": true
   },
   "extends": "eslint:recommended",
@@ -202,7 +203,10 @@
     "padded-blocks": ["error", "never"],
     "padding-line-between-statements": "off",
     "quote-props": ["error", "consistent-as-needed"],
-    "quotes": ["error", "single"],
+    "quotes": ["error", "single", {
+            "allowTemplateLiterals": true,
+            "avoidEscape": true
+    }],
     "require-jsdoc": "off",
     "semi": ["error", "always"],
     "semi-spacing": ["error", {
--- a/src/frontend/js/page-timing.js
+++ b/src/frontend/js/page-timing.js
@@ -144,12 +144,17 @@ PageTiming.reportEndTime = function() {
     if (!localStorage['pageEndTime.' + PageTiming.eventId]) {
         PageTiming.recordInactiveTime();
         var duration = Date.now() - PageTiming.pageStartTime;
+        var activeTime = duration - PageTiming.totalInactiveTime;
+        if (activeTime < 0) {
+            activeTime = 0;
+            console.log("Page tracking: reportEndTime() detected negative activeDuration");
+        }
         var data = {
             type: 'PT',
             eventId: PageTiming.eventId,
             loadTime: PageTiming.pageloadTime,
             duration: duration,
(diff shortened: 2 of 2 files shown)


klown:
@bgoldowsky As suggested in the JIRA, I've reverted back to setting any negative `active_duration` to zero.

Ready for review.

bgoldowsky:
If I remove the conditional so that BOTH the performanceObserver and the old PageTiming code run, they return quite different values for the page load time (667ms vs 1505ms for a page just now).  Do we know why?

klown:
I did not see that when I did my first experiments with the two APIs.  I always got the same value.

klown:
I have duplicated your results.  I have added a number of `console.log()`s that show the two calculations, and dumps out the entire `PerformanceEntry` for the `PerformanceNavigationTiming`.  That structure is documented here:
https://developer.mozilla.org/en-US/docs/Web/API/PerformanceNavigationTiming

Note that the entry's start time ([`PerformanceEntry.startTime`](https://developer.mozilla.org/en-US/docs/Web/API/PerformanceEntry/startTime)) is always set to zero.

My logs show that [`PerformanceEntry.responseEnd`](https://developer.mozilla.org/en-US/docs/Web/API/PerformanceResourceTiming/responseEnd) is approximately equal to the old calculation `window.performance.timing.responseEnd - window.performance.timing.navigationStart`.

The new code uses [`PerformanceEntry.duration`](https://developer.mozilla.org/en-US/docs/Web/API/PerformanceEntry/duration), which is documented as the difference between the time when the [load](https://developer.mozilla.org/en-US/docs/Web/API/PerformanceNavigationTiming/loadEventEnd) of the current document is complete and the `PerformanceEntry.startTime` (always zero).  I think the equivalent in the old API is [`window.performance.timing.loadEventEnd`](https://developer.mozilla.org/en-US/docs/Web/API/PerformanceTiming/loadEventEnd).  And, `window.performance.timing.loadEventEnd - window.performance.timing.navigationStart` is approximately equal to the `PerformanceEntry.duration`.

I think the question is now: what do we want to designate as the "load time"?  I can reconfigure the observer to use the new APIs properties that correspond to the old way of calculating the time.

Also, I have gone through my notes and the history of this branch, but cannot figure out how I got the two methods to agree in the past.  Good catch on your part!

bgoldowsky:
It's hard to understand the difference between responseEnd and loadEventEnd, but I eventually found this [explanation](https://developer.mozilla.org/en-US/docs/Web/Performance/Navigation_and_resource_timings) with a diagram.   

The bottom line appears to be that responseEnd measures the time until "the last byte has been received from the server" - after which the browser takes some time to parse and display the page. loadEventEnd means all of that processing is complete too.

Both seem totally reasonable - but the latter includes more factors that we cannot control, like the speed of user's CPU and display hardware.  I'd say let's stick with the one that's comparable to the data we've previously collected (especially since that older API is still a fallback).

klown:
>  I'd say let's stick with the one that's comparable to the data we've previously collected

I almost made the change to use `responseEnd` in the `observer` yesterday because otherwise any analysis of the data going forward would show a substantial difference from previous analyses, load times being longer than response times.

And your point about differences in processors, memory, etc. is well taken.

I'll make the change to use `PerformanceEntry.responseEnd`.

PR #245, opened 2022-05-12
CSL-1820 alternative to focus in focus out

klown:
Draft PR to show how using:
```
$(document).ready(function() {
  $(document.body).attr('tabindex', '-1');
  $(document.body).focus();
  ...
}
```
allows the use of the `focusin/focusout` machinery to track inactivity on a page.  This is a patch to make that technique work on Chrome.

See:  https://github.com/cast-org/clusive/blob/421f6e67964c99017a08f0f9e9646f7775def83d/src/frontend/js/page-timing.js#L264

(Aside:  I based this on the beacon work to avoid merge conflicts).

JIRA: https://castudl.atlassian.net/browse/CSL-1820


klown:
> The only thing that might be beneficial is a sanity check of the `document.activeElement` to see if it is `<body>` or `null`. before moving focus onto the body. 
> ...
> This may help when an onboarding tooltip is shown so that screen readers keep focus on the tooltip, rather than possibly get forced over to the body.

That makes sense @mbrambilla.  The goal of this PR is for cases where nothing has focus after the page is loaded. If some tooltip or other element is the intended focus target when the page is loaded and focus is moved there, then a focusin event will be emitted and that should suffice.  I'll make the change.

Does anyone know of a page I can test where it is known that focus is set to some element other than the `body`, on page load?

klown:
I wrote:

> Does anyone know of a page I can test where it is known that focus is set to some element other than the body, on page load?

Never mind, I found a case where a tooltip popped up on page load, and the focus went to it rather than the body, and the `focusin/focusout` machinery works in Chrome.  I still need to test in FF and Safari before pushing.

PR #263, opened 2022-06-15
CSL-1428: Library>Delete reading giving 500 error

klown:
Code changes to remove database constraint on `AffectiveCheckResponse` and `ComprehensionCheckResponse` assessment models where their `book` foreign key was set to `PERSIST` when the book record was deleted.  The change sets the deletion constraint to `CASCADE`, meaning the assessment records associated with a specific book will be deleted when the book is deleted.  The actual change is made to the `CheckResponse` model, which is the basis for both of the `AffectiveCheckResponse` and `ComprehensionCheckResponse` models.

Note that this might also fix [CSL-1689: Popular reading> On deleting the article 500 error occurred](https://castudl.atlassian.net/browse/CSL-1689) 

JIRA: https://castudl.atlassian.net/browse/CSL-1428


klown:
@bgoldowsky Ready for review.

PR #267, opened 2022-06-24
CSL-1872: Improve load times

klown:
This adds a `make_positions_and_weight()` function that is used during book import that adds content length to the reading order items in Clusive's `manifest.json`, and generates the `positions.json` and `weight.json` files.

A problem was found with two of the epubs from Clusive's public content.  One was the `tocqueville_democracy-vol1.pub`, in the `gutenberg-tocqueville-democracy-vol1` folder.  The other was volume two.  Volume one's meta-data, and consequently the manifest's reading order, include a link to `../_public_vhost_g_gutenberg_html_files_815_815-h_815-h-19_appendix.xhtml`.  But there is no such entry in the zipfile.

In that case, the `contentLength` for the reading order item in the manifest is left out, and no position nor weight is calculated for it.  I'm not sure if that is the correct thing to do.  The alternative is set its `contentLength` to zero, and put something in for the position and weight (a zero weight?)

Also, `make_positions_and_weight()` does not test for the `publication.metadata.rendition.layout == 'fixed'` condition.  Still to come...

JIRA: https://castudl.atlassian.net/browse/CSL-1872

### DIFF
--- a/src/library/parsing.py
+++ b/src/library/parsing.py
@@ -71,7 +71,7 @@ def unpack_epub_file(clusive_user, file, book=None, sort_order=0, omit_filename=
     This method will:
      * unzip the file into the user media area
      * find the most basic metadata
-     * create the manifest.json
+     * create the manifest.json, positions.json, and weight.json files
      * make a database record
 
     It does NOT look for glossary words or parse the text content for vocabulary lists,
@@ -322,9 +322,15 @@ def make_toc_item(epub, it):
 
 def make_positions_and_weight(epub: Epub, zip_file: ZipFile, manifest: dict):
     """
+    Calculate the positions within and weight of "chapters" in the
+    `readingOrder` section of the `manifest`.  This uses the compressed size of
+    the chapters in the calculations.
+    Note: The `epub` document is passed to determine if it has a fixed layout
+    and to alter the calculations accordingly.  It is currently unused.
     :param epub: EPUB file as parsed by Dawn.
     :param zip_file: the EPUB file packaged in a zipfile
-    :manifest: JSON (dict) created by make_manifest() above.
+    :param manifest: JSON (dict) created by make_manifest() above.
+    :return: a tuple of position_list and weight JSON objects.
     """
     POSITION_LENGTH = 1024
     start_position = 0
@@ -339,7 +345,7 @@ def make_positions_and_weight(epub: Epub, zip_file: ZipFile, manifest: dict):
     for link in manifest['readingOrder']:
         try:
             zip_entry = zip_file.getinfo(link['href'])
-            entry_length = zip_entry.file_size # or compress_size?
+            entry_length = zip_entry.compress_size
             link['contentLength'] = entry_length
             total_content_length += entry_length
             position_count = max(1, math.ceil(entry_length / POSITION_LENGTH))
--- a/src/library/parsing.py
+++ b/src/library/parsing.py
@@ -1,5 +1,6 @@
 import json
 import logging
+import math
 import os
 import posixpath
 import shutil
@@ -200,8 +201,11 @@ def unpack_epub_file(clusive_user, file, book=None, sort_order=0, omit_filename=
         os.makedirs(dir)
         with ZipFile(file) as zf:
             zf.extractall(path=dir)
-        with open(os.path.join(dir, 'manifest.json'), 'w') as mf:
-            mf.write(json.dumps(manifest, indent=4))
+            position_list, weight = make_positions_and_weight(upload, zf, manifest)
+
+        for json_file in (manifest, 'manifest.json'), (position_list, 'positions.json'), (weight, 'weight.json'):
+            with open(os.path.join(dir, json_file[1]), 'w') as jf:
+                jf.write(json.dumps(json_file[0], indent=4))
 
         # If the cover image was retrieved and stored in a tmp file, move it to
         # the new EPUB storage directory.
@@ -211,6 +215,11 @@ def unpack_epub_file(clusive_user, file, book=None, sort_order=0, omit_filename=
         logger.debug("Unpacked epub into %s", dir)
         return book_version, True
 
+
+def save_json_files(dir, file_names, dicts):
+    for file_name in file_names:
+        with open(os.path.join(dir, file_name), 'w') as json_file:
+            json_file.write(json.dumps())
 
 def get_metadata_item(book, name):
     item = book.meta.get(name)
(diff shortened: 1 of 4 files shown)


src/library/parsing.py
```diff
@@ -212,6 +216,11 @@ def unpack_epub_file(clusive_user, file, book=None, sort_order=0, omit_filename=
         return book_version, True
 
 
+def save_json_files(dir, file_names, dicts):
+    for file_name in file_names:
+        with open(os.path.join(dir, file_name), 'w') as json_file:
+            json_file.write(json.dumps())
+
```
bgoldowsky, line 223: Looks like this is not used?

bgoldowsky (commented):
This is looking great!  I think the remaining things to do are
* Check with Aferdita whether it should be file size or compressed file size (I wasn't quite clear on what she said on the call)
* Add pointers to these files into the configuration we send Readium in reader.html
* Test that it prevents the loading of all chapters when the book is opened, and still generates proper positions

src/library/parsing.py
klown: Right, not used.  It was a path I abandoned.  I'll remove it.

klown:
> This is looking great! I think the remaining things to do are
> 
>    - Check with Aferdita whether it should be file size or compressed file size (I wasn't quite clear on what she said on the call)
> ...

I wasn't clear about that either, and I wonder if it matters, as long as one or the other is used consistently.  It wasn't recorded in [the architecture notes](https://castudl.atlassian.net/wiki/spaces/CISL/pages/1969389573/2022-06-21+Architecture+Meeting+notes).  The TypeScript uses `getContentBytesLength()`, which sounds like the uncompressed size, whereas the PHP script uses `comp_size`, which sounds like the compressed size.

@aferditamuriqi, can you clarify?  Thanks!

bgoldowsky:
And regarding Tocqueville, ignoring listed files that don't actually exist sounds like the right thing.  We should fix the file as well since it's clearly invalid. And I think we can put off dealing with fixed layout for another time.

klown:
> > This is looking great! I think the remaining things to do are
> > ...
> > * Add pointers to these files into the configuration we send Readium in reader.html
> 
> There are two new urls in the `readerOptions` structure, namely, `positionsUrl` and `weightUrl`. They have the same form as the `url` for the `manifest.json` file, with the exception of the file names themselves (`positions.json` and `weight.json` respectively).
> 

@bgoldowsky,  @aferditamuriqi,

I looked at the R2D2BC source and saw that the `readerOptions` in Clusive's `reader.html` template is used to initialize the reader, and, specifically, the initializer looks for a `services` block containing `positions` and `weight` members.  The structure of the latter are, afaict, ` { href: URL }`.  Also, the `readerOptions` has a `rights` block that contains a boolean `autoGeneratePositions`.

The latest push first checks if there are pre-calculated `positions.json` and `weight.json` files.  If so, it sets `readerOptions.rights.autoGeneratePositions` to `false`, and adds a services block, like so:

```
readerOptions: {
    ...
    rights: {
        ...
        autoGeneratePositions: false,
        ...
    },
    ...
    services: {
        positions: { href: URL },
        weight: { href: URL }
    }
```

If there are no positions nor weight json files, it set `readerOptions.rights.autoGeneratePositions` to `true` and does not add a `services` block.

Preliminary debugging with a book with pre-calculated positions and weight files shows the correct structure of the `readerOptions`, and the browser's network log shows two successful GET requests for them.

klown:
> I think the remaining things to do are
> ...
>     * Test that it prevents the loading of all chapters when the book is opened, and still generates proper positions

I have two versions of "Alice in Wonderland", one with pre-calculated positions and weight, the other without.  When opening the former, the server logs show that only the chapter viewed is loaded.  In contrast, when opening the other book where R2D2BC calculates positions and weight, the server log shows loading of all of chapters.

Paging through both books appears to be the same, however, it would be good to compare the pre-calculated positions and weight against the ones that R2D2BC auto-generates.

klown:
@bgoldowsky, @aferditamuriqi,

After thinking about it, I reverted the calculations back to using uncompressed file sizes.  Here's why.

As part of the import, Clusive unpacks the epub zip file, stores the epub in an uncompressed form, and deletes the zip file.  Later, when the user chooses a book to read and launches the reader, Clusive serves up the uncompressed form.  The positions/weight calculations auto-generated by the reader use uncompressed sizes since the compressed sizes are no longer available.

I think, then, if the positions/weights are to be pre-calculated during book import, they should do what the reader does and use the uncompressed file sizes.

klown:
@bgoldowsky @aferdita,

A followup to the [comment above](https://github.com/cast-org/clusive/pull/267#issuecomment-1171688766) about the discrepancy in the calculation of the `remainingPositions` property.

The `remainingPositions` property depends on the prior calculation of a `positionIndex` value.  It is set in R2D2BC's `Publication.autoGeneratePositions()` at [line 329](https://github.com/d-i-t-a/R2D2BC/blob/production/src/model/Publication.ts#L329) (note the conditional):

```
const positionIndex = Math.ceil(
        locator.locations.progression
          ? locator.locations.progression
          : 0 * (resource.length - 1)
      );
```
The example code in the [Calculating Positions and Weight](https://docs.google.com/document/d/1eW3HTL_DQ2zdwM8l69FuHfERpuIHPpL4wwG9Mlmi7DQ/edit#) Google doc does not have the conditional.  The TypeScript example there is:

```
let positionIndex = Math.ceil(
    locator.locations.progression * (resource.length - 1)
  );
```

I've written the Python similarly to the latter:
```
position_index = math.ceil(progression * (num_resources - 1))
```

With that background, here is why there is a difference between the two.  The `progression` value is in the range `0.0 - 1.0`, representing a proportion. In the example code (and the Python code), it is scaled by the number of resources and then rounded up to the nearest integer.  As such, the `positionIndex` is in the integer range `[0, 1, 2, 3, ..., n-1]`, where n is number of resources.

In the R2D2BC code, the conditional has the following effect:  if there is any `progression` value, it is rounded up to the nearest integer.  Otherwise, if the `progression` is zero or absent, the `positionIndex` is set to zero.  That entails that once the `progression` is greater than zero, the `positionIndex` is always `1`, since the ceiling of a non-zero proportion  is 1.

The result is the `positionIndex` in the R2D2BC code quickly becomes stuck at 1, and the `remainingPositions` do not change for each `locations` update.  That is, the `remainingPositions` count down for the first two `locations` only and are stuck thereafter at the value for the second location, as the `proportion` climbs towards 1.0.

I think there is an error in the R2D2BC code.  If the `progression` is zero, the number of resources is multiplied by zero -- `0 * (resource.length - 1)` -- but I don't think that was the intent.  Perhaps the idea was to avoid multiplying, subtracting, and calculating a ceiling when the `proportion` is zero?  That is, something like the following was intended?

```
const positionIndex = locator.locations.progression ? 
    Math.ceil(locator.locations.progression * (resource.length - 1))
    : 0
);
```

aferditamuriqi:
> const positionIndex = Math.ceil(
>         locator.locations.progression
>           ? locator.locations.progression
>           : 0 * (resource.length - 1)
>       );

@klown i pushed a fix for this in the [Release - 2.0.6-beta.2](https://github.com/d-i-t-a/R2D2BC/releases/tag/2.0.6-beta.2)

PR #270, opened 2022-07-06
NOJIRA - updated INSTALL.md to include node version requirements and …

lmccormack:
…updated grunt point release.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -17,7 +17,7 @@ Note: all steps between **Create/Activitate Virtual Environment** and **Create a
 * [Postgres](https://www.postgresql.org/) 
   * Version 11.5 or later. Used in the deployment configuration.
 * [Node.js](https://nodejs.org/)
-  * Not required to run the Clusive server, but for development it is needed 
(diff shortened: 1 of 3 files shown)


bgoldowsky:
Grunt version should be updated in package.json as well.

klown:
Note that @dependabot has an [auto-update PR](https://github.com/cast-org/clusive/pull/253) for the same version of grunt.

There is one difference in that dependabot updates `package-lock.json` at [line 34](https://github.com/cast-org/clusive/blob/e1228e0be38f63e374316187bfc19fca2d2baf5e/package-lock.json#L34) to `"grunt": "^1.5.3"`  This PR should do the same for its `package-lock.json` [line 34.](https://github.com/cast-org/clusive/blob/84964187c3b9cbd2f466187fc3a38f755370af25/package-lock.json#L34).

Also, this PR has the change in instructions in `INSTALL.md` and therefore is preferred to dependabot's. After this is merged, [PR#253](https://github.com/cast-org/clusive/pull/253) can be closed, unless dependabot automatically closes it out.

PR #275, opened 2022-07-14
CSL-1814: `aria-live` tweaks, new 'Where am i?', simple notification, toc active fixes

mbrambilla:
Kind of went a little feature overboard while working on the `Where am I?` feature and some updates to how `aria-live` is used.

- some tweaks to how `aria-live` regions are used for the onboarding tooltips
- new `alt+w`/`option+w` shortcut to announce where you are in a reading, for pages outside of the reader, such as the dashboard or library, the page title is parsed and used.
- added a super simple notification/toast item that appears centered at the bottom of the window so visual users get the same information as a screen-reader for the 'Where am I?' shortcut.
  - section portion will return a nested listing for the current section
- updated handling of the TOC active indicator marker so that pages with multiple linked items can have their nested status updated accordingly.
  - support for both `scroll` and `paged` modes
    - `scroll` mode uses middle of scrolling section to determine current location
    - `paged` mode uses the currently visible 'page' to determine current location

Notification showing position and section within the reader:
Screenshot 2022-07-14 151325

Showing the updated TOC active indicator with improved handling for nested links within the same page:
Screenshot 2022-07-14 151402

### DIFF
--- a/src/frontend/js/frontend.js
+++ b/src/frontend/js/frontend.js
@@ -498,7 +498,7 @@ function showTooltip(name) {
                 if (tip_control.is(':visible')) {
                     tip_popover.attr({
                         'role': 'status',
-                        'aria-live': 'assertive',
+                        'aria-live': 'polite',
                         'aria-atomic': 'true'
                     });
                     tip_control.CFW_Tooltip('show');
--- a/src/frontend/js/notify.js
+++ b/src/frontend/js/notify.js
@@ -1,0 +1,44 @@
+/* global notify */
+
+// Super simple notification
+(function($) {
+    'use strict';
+
+    var SELECTOR_CONTAINER = '#notifyContainer';
+
+    var CLASS_FADE = 'fade';
+    var CLASS_IN = 'in';
+
+    var notifyModule = function() {
+    };
+
+    notifyModule.prototype = {
+        show : function(msg) {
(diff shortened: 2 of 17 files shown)


mbrambilla:
Should note the location percentage uses the current location based on the entire publication, NOT the section.  In the screenshot it shows 0% because this is right near the start of the publication.

klown:
> Should note the location percentage uses the current location based on the entire publication, NOT the section. In the screenshot it shows 0% because this is right near the start of the publication.

You could give both since the reader positions data structure has those two values.  For example, for the second position for this book:   "You are 0% through The Adventures of Sherlock Holmes; 2% through A Scandal In Bohemia".

Note that the 0% is based on a `totalProgression` of `0.0033003300330033004`, and the 2% is a `progression` equal to `0.019230769230769232`.  For users, rounding to two decimals is probably sufficient.

klown:
Followup:  the positions data structure has urls like `epub/text/a-scandal-in-bohemia.xhtml` for all of the `positions` within the book/chapter.  That is, there are no sub-sections like I – III.  You must be performing some magic to pull out the subsection text. 🪄

bgoldowsky:
Works well for me.  The notification disappears very quickly though - my gut feeling would be to show it for twice as long.

mbrambilla:
Increased show duration from 3 seconds to 5 seconds.

PR #280, opened 2022-07-20
CSL-710: Update to Django 3

klown:
Changes to update to Django 3, specifically Django 3.2.14.

Highlights:
- `settings.py`:
  -  `DEFAULT_AUTO_FIELD` set to `dango.db.models.AutoField`, meaning model/database migrations are not needed,
  -  `X_FRAME_OPTIONS` explicitly set to `SAMEORIGIN` to override Django's default `DENY`, since Clusive uses frames.
- `requirements.txt`:
  - `django` set to `3.2.14`,
  - `django-progressbarupload` forked and its check for Django version fixed; set to @klown's [fork](https://github.com/klown/django-progressbarupload/tree/fix/django-version).
 
@bgoldowsky upon reflection, I think the `django-progressbarupload` fork should exist in CAST's github space, but I don't have permission to create that.

Note that this has been tested locally and with a docker development container, running the unit tests and the cypress tests.  Still pending is testing the docker production container.

JIRA: https://castudl.atlassian.net/browse/CSL-710


klown:
Another case of errors is in the unit tests.  Dropping into a bash shell in the production container and executing `python manage.py test` gives three errors:

> ```
> ======================================================================
> FAIL: test_logout_event (eventlog.tests.EventlogTestCase)
> ----------------------------------------------------------------------
> Traceback (most recent call last):
>   File "/app/eventlog/tests.py", line 54, in test_logout_event
>     self.assertEquals(event.actor, ClusiveUser.objects.all()[0], "Event should list the user")
> AssertionError: <ClusiveUser: Student1> != <ClusiveUser: Student2> : Event should list the user
> 
> ======================================================================
> FAIL: test_session_created (eventlog.tests.EventlogTestCase)
> ----------------------------------------------------------------------
> Traceback (most recent call last):
>   File "/app/eventlog/tests.py", line 22, in test_session_created
>     self.assertEquals(session.user, ClusiveUser.objects.all()[0], "Session should connect to the user")
> AssertionError: <ClusiveUser: Student1> != <ClusiveUser: Student2> : Session should connect to the user
> 
> ======================================================================
> FAIL: test_does_not_change_periods_user_is_not_in (library.tests.ShareFormTestCase)
> ----------------------------------------------------------------------
> Traceback (most recent call last):
>   File "/app/library/tests.py", line 180, in test_does_not_change_periods_user_is_not_in
>     self.assertFormError(response, 'form', 'periods', 'Select a valid choice. 2 is not one of the available choices.')
>   File "/usr/local/lib/python3.9/site-packages/django/test/testcases.py", line 493, in assertFormError
>     self.assertTrue(
> AssertionError: False is not true : The field 'periods' on form 'form' in context 0 does not contain the error 'Select a valid choice. 2 is not one of the available choices.' (actual errors: ['Select a valid choice. 30 is not one of the available choices.'])
> ```
> 
It's not clear why these three test errors occur.

bgoldowsky:
That is odd. I've never tried running the tests inside a docker container. They run fine in my development environment.

Based on the error messages, it would appear that it's connecting to a database that already has existing users and periods in it rather than creating a new, empty test database as the test code assumes it will.

klown:
I forgot to note it, but, again, I also ran the tests inside a docker container using Django 2 and saw the same failures, so they are likely not due to any Django 3 changes.

> They run fine in my development environment.

Yes, they run without failures for me too, both straight off my local file system, and using docker, but not in the production configuration.  The failures occur when running a docker container in the production configuration.

> Based on the error messages, it would appear that it's connecting to a database that already has existing users and periods in it rather than creating a new, empty test database as the test code assumes it will.

I checked for the `default` test database before and after running the tests, and it did not exist. The console log shows both a `Creating test database for alias 'default'...` at the start of the test run and a `Destroying test database for alias 'default'...` at the end.  The fact that the majority of the tests pass implies that the test database was created and initialized appropriately.  I could possibly halt the tests in the middle, and look at test database records for clues, but it's a low priority.

PR #284, opened 2022-07-28
Feature/CSL-1851 Implement revised dashboard panels for assigned/recent/popular readings

klown:
Implemented improvements to the dashboard for teachers/students/guests crossed with assigned/recent/popular readings.

- Refactored the code such that the assigned/recent/popular panels are all part of the `'popular_readings'` tab-panel and the assigned vs. recent vs. popular readings are tab-panes within that panel, with the appropriate tab selected (Assigned, Recent, or Trending).
- Moved the teacher vs. student vs. guest messages for each of the types of readings to their own template file to be included as appropriate.

JIRA: https://castudl.atlassian.net/browse/CSL-1851
Related: https://castudl.atlassian.net/browse/CSL-1850


src/pages/templates/pages/partial/dashboard_panel_popular_reads_data.html
```diff
@@ -1,32 +1,34 @@
 <ul class="nav box-tabs mb-0">
-    <li class="nav-item"><button type="button" data-clusive-view="popular" class="nav-link popular-read-view-button{% if data.view == 'popular' %} active{% endif %}">Popular reads</button></li>
-    <li class="nav-item"><button type="button" data-clusive-view="assigned" class="nav-link popular-read-view-button{% if data.view == 'assigned' %} active{% endif %}">Assigned reads</button></li>
+    <li class="nav-item"><button type="button" data-clusive-view="assigned" class="nav-link popular-read-view-button{% if data.view == 'assigned' %} active{% endif %}">Assigned</button></li>
+    {% if not is_teacher %}
+    <li class="nav-item"><button type="button" data-clusive-view="recent" class="nav-link popular-read-view-button{% if data.view == 'recent' %} active{% endif %}">Recent</button></li>
+    {% endif %}
+    <li class="nav-item"><button type="button" data-clusive-view="popular" class="nav-link popular-read-view-button{% if data.view == 'popular' %} active{% endif %}">Trending</button></li>
```
bgoldowsky, line 6: "Trending" should be "Popular" for teachers now as well, per design.

src/pages/templates/pages/partial/dashboard_panel_popular_reads.html
```diff
@@ -1,6 +1,10 @@
 <div class="box box-mascot-tipped">
+    {% if is_teacher %}
     <h2>Student reading</h2>
     <p>See what your students are reading and what they said they learned.</p>
```
bgoldowsky, line 4: This message is duplicated now.
image

src/pages/templates/pages/partial/dashboard_panel_popular_reads_student_messages.html
```diff
@@ -0,0 +1,34 @@
+{% if data.view == 'assigned' %}
+    {% if data.all|length == 0 %}
+        Here’s where you’ll see readings assigned by your teacher.
+    {% else %}
+        These are your {{ data.all|length }} most recently assigned readings. See all assigned
```
bgoldowsky, line 5: Currently says "These are your 1 most..." if there is only one.
But perhaps rather than do the singular/plural thing we can drop this message/link entirely if there are less than 4 readings? It becomes useless.

src/pages/views.py
```diff
@@ -265,15 +252,53 @@ def get_context_data(self, **kwargs):
...
 
 
+def get_recent_reads_data(clusive_user):
+    return {
+        'view': 'recent',
+        'all': Paradata.latest_for_user(clusive_user)
+    }
+
 def get_popular_reads_data(clusive_user, periods, current_period, assigned_only):
     # Gather data about books that are trending for dashboard view
```
bgoldowsky, line 303: The 'assigned' teacher panel used to be "the most popular readings that are also assigned", but based on the messages we need to change this query.  It is now "the 3 most recently assigned readings". Query needs to be changed to reflect this.

bgoldowsky (changes requested):
Added some comments in the diff.
Two additional things - 
For student/guest, the "Recent" button is throwing an error for me: "Uncaught ReferenceError: target is not defined".
I do not think the 'popular' listing for guests is working.

src/pages/templates/pages/partial/dashboard_panel_popular_reads_data.html
klown: Okay, done.

src/pages/templates/pages/partial/dashboard_panel_popular_reads.html
klown: Thanks for noticing the duplication.  The upper message was missed because `grep` did not find it.  While the gist is the same, the text is actually different. I've deleted it.

src/pages/templates/pages/partial/dashboard_panel_popular_reads_student_messages.html
klown: Part of the problem is that the readings lists are truncated to 3 readings in the code, so there is always fewer than 4 readings by the time the template is processed.  I've added a new property, `total`, that is used by the template.

Also, experimenting with removing the link depending on that total, I found (1) the experience to be inconsistent ("where did the link go?"), and (2) the link was still convenient for navigating to the library showing the assigned readings.  So, I left the message/link in even when there was fewer than 4 items in the total.  But, that is a sample size of one ...  Anyhow, if there are fewer than four, the message leaves out the "3":  "These are your most recently assigned readings ...".

src/pages/views.py
klown: Okay, there is now a new query, `BookAssignment.recent_assigned()`, that retrieves the most recently assigned readings for the period, ordered by date assigned.

But, I'm not sure if this makes `BookTrend.top_assigned()` obsolete?

klown:
> ... Two additional things 
> For student/guest, the "Recent" button is throwing an error for me: "Uncaught ReferenceError: target is not defined".

I found a bug, but it led to a different exception: `TypeError: unhashable type: 'slice'`.  I've fixed that; hopefully it fixes the `ReferenceError`.  (I did not see a `ReferenceError`).

> I do not think the 'popular' listing for guests is working.

You're right.  I found a couple of issues here -- 'popular' was not working correctly for students either -- and they are fixed.  Part of the fix involved modifying `BookTrend.top_trends()` to handle the case where there is no `period` context. In that case it retrieves the most popular readings across all periods.  The intent is to capture the most popular readings among all Clusive users.

And, a couple of additional issues:
- The partial templates `dashboard_panel_popular_reads.html` and `dashboard_panel_recent_reads.html` are no longer used.  Should they be deleted?
- The library cards for the "Recent" tab lack some information because the readings list is based solely on `Paradata.latest_for_user()`.   The cards do not have customizations, starred, etc.  Now, that is how it is done in the current release so perhaps it's okay.  But should they have more information, such as customizations, etc.?  Maybe a new JIRA?

bgoldowsky:
Yes, please remove unused templates.
Students don't see customizations so that doesn't need to be added, but stars should show up. Stars are part of Paradata so that shouldn't be hard?

src/pages/templates/pages/partial/dashboard_panel_popular_reads_guest_messages.html
```diff
@@ -0,0 +1,23 @@
+{% if data.view == 'assigned' %}
+    A list of recently assigned readings displays for students with a
+    Clusive account connected with a teacher.
+{% elif data.view == 'recent' %}
+    {% if data.all|length == 0 %}
+        You have no recent readings yet. Go to the 
+        <a href="{% url 'library_style_redirect' view='period' %}">library</a>
```
bgoldowsky, line 7: Should probably be view='public' for guests, since they have no period.

src/pages/templates/pages/partial/dashboard_panel_popular_reads_guest_messages.html
```diff
@@ -0,0 +1,23 @@
...
+        and get started in Clusive!
+    {% else %}
+        Revisit something you've read, or head to the
+        <a href="{% url 'library_style_redirect' view='period' %}">library</a>
```
bgoldowsky, line 11: Also 'public'

src/pages/templates/pages/partial/dashboard_panel_popular_reads_guest_messages.html
```diff
@@ -0,0 +1,23 @@
...
+        for something new.
+    {% endif %}    
+{% elif data.view == 'popular' %}
+    {% if data.all|length == 0 %}
+       There are no readings popular with Clusive readers right now. Once they
+       do more reading, you will find popular titles here.
+    {% else %}
+        These readings are popular with other Clusive readers. See all popular
```
bgoldowsky, line 19: We should drop the 'see all popular readings' link for now, since we don't have a library view for that yet.

src/pages/templates/pages/partial/dashboard_panel_popular_reads_data.html
```diff
@@ -1,32 +1,34 @@
...
-            <p class="mx-1">
-                {% if data.view == 'assigned' %}
-                    Your students haven't done any assigned reading, yet.
+                {% if view == 'recent' %}
+                    {% include "library/partial/library_card.html" with book=item.book show_starred=False show_description=True show_menu=False show_topics=False link_stretch=True %}
+                {% elif data.view == 'assigned' and is_guest %}
+                    {% include "library/partial/library_card.html" with book=item show_starred=False show_description=True show_menu=False show_topics=False link_stretch=True %}
                 {% else %}
-                    Your students haven't done any reading, yet.
+                    {% include "library/partial/library_card.html" with book=item.trend.book unauthorized=item.unauthorized read_count_data=item.trend.user_count graph_data=item.comp_check graph_link_prefix="comp-detail-all" graph_link_serial=forloop.counter custom_link_prefix="custom-detail-all" custom_question_data=item.customization show_starred=True show_description=False show_topics=False show_assignments=True show_menu=True %}
```
bgoldowsky, line 28: These parameters include "teacherly" items in the student view.  
For students, remove the read_count_data, graph_*, custom_*, and show_assignments; and set show_description to true.

src/pages/views.py
```diff
@@ -265,21 +252,63 @@ def get_context_data(self, **kwargs):
...
-            'data': get_popular_reads_data(self.clusive_user, periods, current_period, assigned_only=assigned),
+            'data': readings,
         })
         return context
 
 
+def get_recent_reads_data(clusive_user):
+    return {
+        'view': 'recent',
+        'total': Paradata.latest_for_user(clusive_user),
```
bgoldowsky, line 299: I think 'all' should be set here, and 'total' set to a number. As is, 'recent' view is always empty.

bgoldowsky (changes requested):

src/pages/views.py
bgoldowsky: I think this function actually returns the _least recently_ assigned!

klown:
> Yes, please remove unused templates.

Okay, but my mistake regarding one of the two I cited.  It should have been `dashboard_panel_getting_started.html`, not `dashboard_panel_popular_reads.html`.

> Students don't see customizations so that doesn't need to be added, but stars should show up. Stars are part of Paradata so that shouldn't be hard?

Yes, it's easy since it is available from the Paradata, and it involves changing a parameter on the` {% include ... %}` directive.  Done.

I also found an error due to a copy/paste/undo which broke the recent readings query that [you also noticed](https://github.com/cast-org/clusive/pull/284#discussion_r938216795).  It's fixed.

Moving on to the latest set of comments.

src/pages/templates/pages/partial/dashboard_panel_popular_reads_guest_messages.html
klown: I was wondering about that since whenever it was clicked, the library page showed some list other than a list of popular books.  Okay, I'll drop it.

src/pages/views.py
klown: Oops.  Fixed.

src/pages/templates/pages/partial/dashboard_panel_popular_reads_guest_messages.html
klown: Right.

src/pages/templates/pages/partial/dashboard_panel_popular_reads_guest_messages.html
klown: Done.

src/pages/templates/pages/partial/dashboard_panel_popular_reads_data.html
klown: Okay, done.  However, I find showing the period assignment useful since, for students, there is no indication of the period on the dashboard. But its is not in the cards in the current release, so I've removed it.

klown:
@bgoldowsky , This should handle the issues we talked about on Slack last week -- excluding unauthorized readings from the "Popular" tab-panel and showing user counts in the "Assigned" panel for teachers.

I looked into QuerySets and filters for that functionality but did not see how to do it, and ended up with a couple of functions to produce the results.

Ready for review.

PR #286, opened 2022-08-05
CSL-1930: Mockup tour popovers, chaining code

mbrambilla:
Base layout and styling for tour popovers, also referred to as the 'tooltip' tour.

Contains basic chaining functionality to move between popovers through the use of CSS selectors.  The selectors used in the function calls refer to the trigger element even though the triggers will not actually invoke the show/hide of the popover. The 'trigger' will be where the 'arrow' portion of the popover will point to.  

Currently each popover needs to be initialized through the `tour.prepare(selector)`.

A 'start tour' button will be used to begin the tour, and movement through the chain will be controlled through links in each popover, using `tour.chain(selector)`.   Included is a a customized close routine, `tour.closeRefocus(selector)` to close the popover tour and return focus to a specified element, most likely the 'start tour' button.

How the chain gets put together is potentially a sticking point as it could be contextual based on page, user role, visible panel (?), and possibly other parameters.  Depending on the featured item on the page a custom element may be needed as a hidden trigger to get proper alignment of the popover.


bgoldowsky:
Testing on Firefox/Mac:
* The "1/3" label chages to 2/3 but never to 3/3 in my testing.
* I can only back up one step with the "Prev" button, not all the way back to step 1.

mbrambilla:
Typos have been fixed.

PR #290, opened 2022-08-11
CSL-1910: Modify library page for reading level (text difficulty) filters

klown:
- [x] Add `reading_level` field to `BookVersion` model,
- [x] Add "Reading Level" checkboxes panel,
- [x] Add reading level badges to library cards,
  - Uses the labels as given in [CSL-1909](https://castudl.atlassian.net/browse/CSL-1909?focusedCommentId=35143).
- [x] Implement filtering function for reading level checkboxes.
- [x] Implement filtering function for library card reading level badges.

JIRA: https://castudl.atlassian.net/browse/CSL-1910


klown:
This code in the [latest push](https://github.com/cast-org/clusive/pull/290#commits-pushed-547adfc) is still rough and needs work.  In particular:

1. While the switch to using integers for reading levels is done, filtering  by reading levels still uses `BookVersion` to find the books that fit the filters.  I will keep looking into how to use ranges at the `Book` level to do the same thing.  The first attempt did not work.
2. I have not used the ARI work nor checked that my changes work with it.
3. There may be one too many migrations now. Or it's likely cleaner to remove the first migration and keep the last one.
4. There is likely code that was written for the enumerated category reading level approach that no longer used. Some clean up is needed.

klown:
The latest commit addresses item one in the previous comment:

> 1. While the switch to using integers for reading levels is done, filtering  by reading levels still uses `BookVersion` to find the books that fit the filters.  I will keep looking into how to use ranges at the `Book` level to do the same thing.  The first attempt did not work.

Now the `Book.min_reading_level` and `Book.max_reading_level` are used to exclude books outside of the range defined by the reading level checkbox filters.

However, this assumes that the reading levels for a book are a contiguous set.  But, I don't _think_ that is required.  For example, a book might have a version that is "Elementary" reading level (level 2), and another version that is "High school" (level 4).  Using  the `Book`'s min and max results in erroneously showing that book's library card when only the "Middle school" (level 3) filter is checked.  That's because "3" is in the range 2--4, even when there is no such version of the book.

This suggests that [using](https://github.com/cast-org/clusive/pull/290/commits/5812ce976682751f21fbddefe05b7a38c44d3dff) the `BookVersion.reading_level` for the filtering is more accurate since the example book would not be included.  There is no `BookVersion` with a reading level of 3 that can be included, and the card for that book will not be shown.

Update: the converse is also true when the filtered range is not contiguous.  Suppose, for example, the "Early reader" (level 1) and "Middle school" (level 3) filter checkboxes are the only ones checked, defining a range of 1--3.  A book that has only a version with a reading level of 2 is within that range, and its library card shows, even though the "Elementary school" (level 2) is **not** checked.  Abstractly, the procedure should be to check each filter against each book reading level and show only matches.  I believe the only way to do that is to base the query on the `BookVersion`.

klown:
The [just previous commit](https://github.com/cast-org/clusive/pull/290/commits/87f4afbaeb2358ee4accedd07b8240b82e79c77f) addresses the last item in [the comment from above](https://github.com/cast-org/clusive/pull/290#issuecomment-1215056831):

> 4. There is likely code that was written for the enumerated category reading
> level approach that no longer used. Some clean up is needed.

Done.

klown:
Thanks @mbrambilla for the fixes to the layout.

@bgoldowsky This is far enough along that I've marked it "ready for review".

However, somehow the activation state and checkbox state for "Topics" and "Word Count" are no longer working.  Their filtering operation works fine.  Selecting topic(s), or word count(s) does the correct filtering in terms of displaying library cards.  However, the `active` styling of the relevant library card tags is not applied; neither is the `checked` state of the relevant checkbox filters.

I'll see if I can find the problem, although I thought I had left that code untouched.

klown:
> However, somehow the activation state and checkbox state for "Topics" and "Word Count" are no longer working.
>  ...
> I'll see if I can find the problem, although I thought I had left that code untouched.

FIxed.  Likely due to a copy/paste error that added an extra comma.

PR #294, opened 2022-08-19
Address CSL-1851 dashboard QA comments

klown:
This PR addresses both changes to the messages and the duplicate titles for a guest user when viewing the "Popular" panel.

The message modifications where based on [comment 5288](https://castudl.atlassian.net/browse/CSL-1851?focusedCommentId=35288) in the JIRA.

Regarding the duplicate books issue, [Django QuerySets](https://docs.djangoproject.com/en/3.2/ref/models/querysets/#django.db.models.query.QuerySet.distinct) support a`distinct(fieldName)` method that could be used to remove duplicates from the book trends, i.e. `distinct('book')`.  But:
1. `distinct(fieldName)` is supported by PostgreSQL only, and
2. `order_by('book')` must be used in conjunction with `distinct('book')`.  But, `order_by('-popularity')` is what is desired here.  It might be possible to order by popularity after the duplicate books are removed from the set; not sure.

While there may be a way to use `distinct()` in our case, for now I have implemented a function that removes duplicated books from the set of book trends.  It does the job and works with SQLite.

JIRA: https://castudl.atlassian.net/browse/CSL-1851

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -12,16 +12,17 @@ Note: all steps between **Create/Activitate Virtual Environment** and **Create a
 
 * [Python 3](https://www.python.org/downloads/)
   * Version 3.7.4 up to 3.9. (3.10+ not tested yet). On Mac, Homebrew is the easiest way to install.
-* Django 3
-  * Version 3.2.14+, installed with other Python dependencies as documented in
+* Django
+  * Version 3.2, installed with other Python dependencies as documented in
     section [Install Python Dependencies](#install-python-dependencies).
 * [virtualenv](https://virtualenv.pypa.io/en/latest/) 
-  * Not required, but highly recommended for maintaining an isolated environment and dependencies.
+  * Not required, but highly recommended for maintaining an isolated environment for Python and Python packages.
 * [Postgres](https://www.postgresql.org/) 
   * Version 11.5 or later. Used in the deployment configuration.
 * [Node.js](https://nodejs.org/)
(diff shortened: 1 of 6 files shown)


klown:
@bgoldowsky Ready for review.

bgoldowsky:
So, this "trending" view for guest is based on "local" popularity - in the sense of books that are extremely popular within some single classroom - rather than what you might call "global" popularity.  E.g., If one student in each of 100 classes read a book yesterday, it would lose out in this ranking to a book read by 5 students in a single class.

I don't think that's "right", but I do think it is "good enough" given that this is just supposed to give guest users a sense of what the dashboard display is like. If we want registered users to have a "globally trending" view in addition to the "classroom trending" views, then we would need to implement that (probably with a new summary table since it would be expensive to calculate).

klown:
@bgoldowsky wrote:

> I don't think that's "right", but I do think it is "good enough" given that this is just supposed to give guest users a sense of what the dashboard display is like.

That's a good point about a book's overall popularity independent of the periods in which it's popular.  A note, however, just in case this should be addressed sooner rather than later.  This also applies to any student user who has not been assigned to any periods.  They see the same popular readings as a guest would, since no period is in effect on their dashboard. And, the same occurs for teachers who haven't created any periods yet -- again, because there is no period context for the dashboard.

I suspect the teacher case is relatively transient -- they will create periods.  And, students will likely be assigned to some period before long.

PR #307, opened 2022-09-13
CSL-1998: Persistent tab choice for dashboard 'popular reads' panel

mbrambilla:


bgoldowsky:
Why did you switch from the explicit list of choices you had for the model field in your first commit?  Having that list was more consistent with the way we've done other similar choices the user makes (library_style, transform_tool, etc).

mbrambilla:
The idea was to reduce the number of locations that needed specifically list out what the panels are and keep the master list within the Python code. Having to update Python, templates, and DB just to add/remove a panel type seemed excessive to me.

It may help deal with the possibility of changing an account type between teacher/student or removing a panel type later with the fall through in `get_readings_data`.  The DB entries will not need to be specifically updated if a panel type is removed or if the default panel type is changed.

bgoldowsky:
The choices lists are not actually stored in the DB (even though they show up in the migrations files) - the validity of values being stored is just checked in the django model code.  But I hear what you are saying about keeping things flexible.

PR #308, opened 2022-09-13
CSL-1950 Limit unsuccessful logon attempts

klown:
- [x] Make use of `django-axes` cool off time,
- [x] Integrate with Clusive's forgot-my-password workflow,
  - Temporary unblock for password sent by Clusvie to user’s email and, if successful, unblock that user,
- [x] Add countdown warnings when user is close to their failure limit
  - Example:  “You have 3 attempts remaining”.

JIRA: https://castudl.atlassian.net/browse/CSL-1950


klown:
This is almost ready.  The last changes rely on axes own internal timer to reset the lock when the cool off period ends.  

However, by default, django-axes restarts its cool off timer if the user attempts to login again during cool off.  For example, if cool off is 15 minutes, and the user waits 10 minutes before trying to log in, axes' timer is reset back to 15 minutes.

For now, `settings.py` includes a `AXES_RESET_COOL_OFF_ON_FAILURE_DURING_LOCKOUT = False` to avoid this scenario. This might be a security issue.  If it is and the setting is changed back to the default behaviour, the code that calculates and displays the remaining time to wait will need to be changed.

Also, the retry limit and cool off time are set for development, at 4 retries and 15 minutes respectively.  They should be set to realistic values for production before merging.

klown:
@bgoldowsky, I found a setting for logging all login failures and noticed that one of the fields in that model is whether the failure caused a lockout.  Here is what is stored in the failure records:

- timestamp
- username
- client IP address
- user agent
- path, i.e. `/account/login`
- locked out (boolean)
-  other fields ...

Logging in this way entails having records of all the login failures whether they cause a lockout or not.  But, when processing the logs, one can filter for "locked out" records.  Is that good enough for tracking lockouts?  Or would it be better to incorporate the lockout into Clusive's event log?

Note that the `settings.py` in this PR is configured for logging failures.

klown:
BTW, in addition to this code there is some setup required in order to use django-axes.  Many of the steps in the installation document can be ignored as they are covered by the modifications to `settings.py`.  But, these three steps are still needed.

1. `$ pip install django-axes` ([pre step 1](https://django-axes.readthedocs.io/en/latest/2_installation.html#installation))*
2. `python manage.py check` (step 4)
3. `python manage.py migrate` (step 5 -- install django-axes models/tables)

\* Should django-axes be added to `requirements.txt`?

klown:
@bgoldowsky I've addressed your and Kim's comments on the JIRA regarding the warning messages:  lower case and bold "Warning", styling the colour of the change-your-password link to match the rest of the text, and changing the wait message to show the time remaining as HH:MM.

klown:
As per yesterday's technical meeting, split the difference and set the maximum number attempts to 15.

PR #317, opened 2022-09-26
CSL-1997 Add text complexity to the upload/edit metadata page

klown:
Modifies the base `MetadataFormView` and associated form, urls, and derived views to allow the user to modify the reading difficulty level of a book, specifically a version (aka `BookVersion`) of that book, and save the new value.

JIRAs:
- https://castudl.atlassian.net/browse/CSL-1997
- https://castudl.atlassian.net/browse/CSL-1911


klown:
@bgoldowsky the code is ready for review, however, I ran into two issues.

Regarding the main task -- to add radio buttons for setting the reading level during an upload or edit -- it's necessary to pass the `BookVersion` id in the urls since the reading level is stored in a `BookVersion`, not in a `Book`.  In all cases but one, the `BookVersion` id is readily available and can be added to the url.  The one case where that is not possible is the "Edit" menu item in the dropdown menu on a library card.  Only the `Book` id is available in that case, and I could not see a way to get a specific `BookVersion`.  For now, the default is to use the first `BookVersion` associated with that `Book`.

Secondly, I had to make a change to the `ReadingLevel` class with respect to an internal comparison to an incoming grade. Specifically, I ran into an exception with the `ReadingLevel.for_grade()` class function where the `grade` was 14.  That should return `ReadingLevel.ADVANCED` , but an exception was raised instead.  The cause:  there is a comparison within `for_grade()`:

> `if grade <= level.max_grade: ...`

For `ReadingLevel.ADVANCED`, the maximum grade was `None`. But it's illegal to compare an integer with `None`; hence the exception.

To fix it, I changed the maximum grade for `ReadingLevel.ADVANCED` to `math.inf` (positive infinity).  Similarly I changed the minimum `ReadingLevel.EARLY_READER` to `-math.inf` (negative infinity).   I also added unit tests to make sure this works; and searched the code for any cases where there was a reliance on the previous `None` value.  I found only two, both in the `LibraryDataView` reading level filters where there are checks for advanced and early readers.  I swapped out the `None` with the appropriate type of infinity.

PR #318, opened 2022-09-27
CSL-1998: Fix test case by setting value for dashboard panel

klown:
@bgoldowsky, @mbrambilla,

The roster `test_manual_anon_id()` failure is due to the `ClusiveUser` test instance where the `dashboard_popular_view` is not set.  Now, the test is actually checking whether the clusive user's anonymous id can be set manually, but. after setting the anonymous id, the `full_clean()` check fails with a `ValidationError: {'dashboard_popular_view': ['This field cannot be blank.']}`.

A simple fix is to set the `dashboard_popular_view` before the `full_clean()` check.  This PR sets it to 'assigned'.

The [default value](https://github.com/cast-org/clusive/blob/development/src/roster/models.py#L239) in the ClusiveUser model for the `dashboard_popular_view` is an empty string, and I do not see any other restriction on it, in particular,a restriction that it "cannot be blank".

Also puzzling is the next test, `test_anon_id_unique_enforcement()` that also uses a `full_clean()` check.  But, this test passes even though the `dashboard_popular_view` is not set there either.


klown:
I just figured out why the `test_anon_id_unique_enforcement()` passes.  It's set up to generate a `ValidationError` due to the duplicate anonymous id's, which it does.  Examining the full set of errors shows `dashboard_popular_view: ['This field cannot be blank.']` as another validation error.

klown:
I see what is going on -- when declaring a model field, the default value of `blank` is `False`.  So another way to fix this is to declare that field with `blank=True`.  Unless, of course, a blank field is not allowed here.

mbrambilla:
Yeah, my bad.  
Confirmed that changing the model to:
`dashboard_popular_view = models.CharField(max_length=10, blank=True, default="")` does resolve the test error.

The idea was to use `blank` as the default state in the DB and let Django figure out what the default panel should be in code, since it could possibly vary by role, since and anonymous user should not have any 'assigned' readings.

klown:
Thanks @mbrambilla.  In that case, here is the change to the model to allow a blank `dashboard_popular_view`, and the accompanying migration file.  (And, resetting the test.py code back to what it was).

PR #328, opened 2022-10-14
Feature/CSL-2007/CSL-2040 group 1 and 2 enhanced tooltips

klown:
Create the "group 2" popover dialogs and define their targets such that they show when the relevant page is loaded, and the criteria for showing them is met.

Note that since this PR is an extension of the group 1 popovers, it is based on and includes the changes in PR #323.  Going forward this PR will track changes for both group 1 and group 2 enhanced tooltips.

JIRAs:
- https://castudl.atlassian.net/browse/CSL-2040
- https://castudl.atlassian.net/browse/CSL-2007


klown:
@bgoldowsky , while I've marked this as "ready for review", there are changes to both the kind of tip for Cluey, and its content.  It looks like it will revert to an actual tooltip that is shown on mouse hover, and not a spontaneous display that depends on the history of when it was last shown.  See [Kristen's comment](https://castudl.atlassian.net/browse/CSL-2040?focusedCommentId=36795) in the JIRA.

Given that change, I need to at least change the logic for showing Cluey's popover, likely remove it from that logic.

klown:
Based on a conversation with Kristen in the JIRA for CSL-2007, the group 2 [Active Reading time and Details](https://docs.google.com/document/d/1Pod1qiY8agv86j-SMv0pzkdzwYr4KfdjWJ-eDagQNjM/edit#heading=h.6c6bur4n6pdv) is a duplicate of the group 1 [Tooltip_student activity](https://docs.google.com/document/d/1r9sXmIIbpsuMZmpa5iqvw_FoEHSyDxO5YbbSd3VRp1I/edit?usp=sharing).  The latter is a replacement for an existing tooltip and implemented in PR #323.  The former was implemented in this PR as a new popover.  I will remove the new popover from here.  This entails removing a `TipType` from the database and adjusting priorities as appropriate.

The [conversation in the JIRA](https://castudl.atlassian.net/browse/CSL-2007?focusedCommentId=36823) starts with a note that the group 1 popover is missing from the Dashboard tour.

klown:
@mbrambilla FWIW, after the meeting on Wed afternoon, I put the [page-based lists of the different popover types](https://github.com/cast-org/clusive/blob/cac2e3d909f6e97d9165ee26b6f7f6d4c6ae9925/src/tips/models.py#L21) in tour-order.  That order does not matter for single automatic onboarding display of the popover, but it might be useful to you.

Regarding the meeting itself, look for the section "Tool tip tour questions" in the [meeting notes](https://castudl.atlassian.net/wiki/spaces/CISL/pages/2055176193/2022-10-19+Meeting+notes).

klown:
@mbrambilla By chance, I noticed that when the video in a popover is playing, closing the popover with either the close button, or the "Got it!" button, the audio track of the video keeps playing.  Actually, I think the video itself keeps playing but it is invisible, so it's just heard.

I think that's because closing the popover just hides it: `$curr.CFW_Popover('hide')`, in [tour.js](https://github.com/cast-org/clusive/blob/feature/CSL-2040-group-2-enhanced-tooltips/src/frontend/js/tour.js#L63).  And that's possibly because the popover is part of a tour, and could be shown again by pressing a "previous" button from the next popover in the tour ?  Keeping it around, but hidden, makes the tour easier to present.

By comparison, the previous tooltips were disposed when closed, after first being hidden: `$(this).CFW_Tooltip('dispose');`  from [frontend.js](https://github.com/cast-org/clusive/blob/development/src/frontend/js/frontend.js#L517).

Back to the popovers:  disposing the popover when closed does indeed stop the video and its sound track.  But, I'm not sure it is the right thing to do, or, perhaps the right thing in all contexts.

Alternatively, and possibly a rabbit's hole, use an API for interacting directly with the video and stop it when the popover is closed (hidden)?

mbrambilla:
@klown Thank you for the reminder that I needed to look into pausing the YouTube videos on hide of the popovers (and modals).

Will look into possibly hooking into the [YouTube iframe API](https://developers.google.com/youtube/iframe_api_reference), or through more nefarious methods such as resetting the `<iframe>`'s `src` attribute as mentioned in a StackOverflow post about [Youtube Video Still Playing When Bootstrap Modal Closes](https://stackoverflow.com/questions/22613303/youtube-video-still-playing-when-bootstrap-modal-closes)

Also reference for the [SproutVideo Javascript API](https://sproutvideo.com/help/articles/27-javascript_player_api).

klown:
@bgoldowsky @mbrambilla This is ready for review. All the group 1 and group 2 "enhanced tooltips" are in place, including the recent two for the student Dashboard.

The last code I added was a couple of unit tests for changes I made to `TipType.can_show()`.  Good thing, since I found and fixed bug in my changes.  I'm inclined to add more automated tests for the logic underlying what/if a popover should show.  What do you think?

As we discussed at the technical meeting, this PR is ready to merge @mbrambilla's code from PR #333 into it.  Alternatively, this could be reviewed and merged for Q/A while the tour work continues on its branch.  Let me know how you think we should proceed.

src/library/views.py
```diff
@@ -389,7 +390,13 @@ def get(self, request, *args, **kwargs):
     def get_context_data(self, **kwargs):
         context = super().get_context_data(**kwargs)
         context['search_form'] = self.search_form
-        context['tip_name'] = self.tip_shown.name if self.tip_shown else None
+        # BEGIN: Sample Tour
```
bgoldowsky: Is this a sample, as opposed to something that should remain for production use?

src/pages/views.py
```diff
@@ -223,7 +223,14 @@ def get_context_data(self, **kwargs):
         context['query'] = None
         context['panels'] = self.panels
         context['data'] = self.data
-        context['tip_name'] = self.tip_shown.name if self.tip_shown else None
+        context['clusive_user'] = self.clusive_user
+        # BEGIN: Sample Tour
+        # Sample tour with single item list
```
bgoldowsky: sample?

src/pages/views.py
```diff
@@ -710,7 +718,13 @@ def get(self, request, *args, **kwargs):
             'annotations': annotationList,
             'cuelist': json.dumps(cuelist),
             'hide_cues': hide_cues,
-            'tip_name': self.tip_shown.name if self.tip_shown else None,
+            # BEGIN: Sample Tour
```
bgoldowsky: sample?

src/tips/models.py
```diff
@@ -24,18 +84,24 @@ class TipType(models.Model):
     def can_show(self, page: str, version_count: int, user: ClusiveUser):
         """Test whether this tip can be shown on a particular page"""
         # Teacher/parent-only tips
-        if user.role == Roles.STUDENT and self.name in ['book_actions', 'activity']:
+        if user.role == Roles.STUDENT and self.name in TEACHER_ONLY_TIPS:
```
bgoldowsky: Also Roles.GUEST or use something like "not user.can_manage_periods".  Currently guests are being assigned teacher-only tips.

src/tips/models.py
```diff
@@ -24,18 +84,24 @@ class TipType(models.Model):
...
             return False
-        # Switch tooltip requires multiple versions
+        # Switch TipType requires multiple versions
         if self.name == 'switch':
             return page == 'Reading' and version_count > 1
+        # Thoughts TipType is only for students
+        if self.name == 'thoughts' and user.role != Roles.STUDENT:
```
bgoldowsky: Also GUEST, see above.

src/shared/templates/shared/partial/popover_tour.html
```diff
@@ -0,0 +1,384 @@
...
+                    {% endif %}
+                {% endif %}
+            </div>
+            <div class="col-auto">
+                <button type="button" class="btn btn-primary" data-cfw-dismiss="popover">Got it!</button>
+            </div>
+        </div>
+        {% elif tour_name == "book_actions" %}
+<!-- book_actions -->
+        {% if clusive_user.role == 'TE' %}
```
bgoldowsky: This file with all of its 'if' statements feels extremely fragile, it would be great to do some refactoring. Why is this one just TE, while most are TE or PA? Is there some reason that parents should not get to see the video?

src/shared/templates/shared/partial/popover_tour.html
```diff
@@ -0,0 +1,384 @@
...
+        <div class="embed-fluid embed-fluid-16x9 mb-0_5">
+            <iframe id="book_actionsIframe" src="https://videos.sproutvideo.com/embed/799edbb41512e5c4f0/ba8fab36c9170dc4?type=hd" frameborder="0" allow="accelerometer; autoplay; clipboard-write; encrypted-media; gyroscope; picture-in-picture" allowfullscreen aria-label="Book actions video"></iframe>
+        </div>
+        {% endif %}
+        <p>Click “More actions” (three dots) to assign or customize any book, and for books you uploaded to Clusive, to edit information or delete a book.</p>
+        <div class="row gx-0_5 flex-items-center">
+            <div class="col">
+                {% if has_teacher_resource %}
```
bgoldowsky: What is the purpose of passing in the 'has_teacher_resource' variable, given that this template has hard-coded the URL of any teacher resource that exists already? In this case, has_teacher_resource is passed in as false by LibraryView. Clearly there is one, but the link is not displayed.

bgoldowsky:
The PR is a little confusing because it contains things related to the tour, but doesn't actually implement the tour. So I think things like the 'tour samples' passed in context variables are harmless, but hope they will be cleaned up in the next PR.

There are defintely some logic errors, mostly due to checking for STUDENT or TEACHER but forgetting about GUEST and PARENT.

src/library/views.py
mbrambilla: This is kind of a hybrid mode between the tooltips (tip_name) and the popovers (tours).   This has been cleaned up in PR #336.  I was originally planning on add the tour chains into the original PR, but things forked in two different directions into this PR and then the tour PR.

src/pages/views.py
mbrambilla: Transitional hybrid mode resulting from trying to do too much in one PR, cleaned up in PR #336.

src/pages/views.py
mbrambilla: Transitional hybrid mode resulting from trying to do too much in one PR, cleaned up in PR #336.

klown: What are the implications for this PR, then?  Is it simply a matter of removing the "sample tour" comments?  It goes back to the issue of whether the singleton popovers can be merged vs. combining this branch with the tour branch.

src/tips/models.py
klown: I should have included a GUEST user in the tests.  When GUEST is tested, the `TipType.can_show()` fails for the `activity` tip on the Dashboard.

By implication, teacher tips should also show for guests using the current version  in production where only `book_actions` and `activity` tips are checked in `TipType.can_show()`.  Do `book_actions` or `activity` tooltip pop-up for guests?  I suspect not because the Dashboard for guests does no include the relevant activity panel nor the book_actions menu.  As such, there is nothing for the tooltip to point to.

But, in the current situation, there are more `TEACHER_ONLY_TIPS` some of whose targets appear for guest users.  So, I'll check for guest users.

src/tips/models.py
klown: Right.

src/shared/templates/shared/partial/popover_tour.html
klown: Yeah, I'll see if I can refactor and simplify the template.  As for the missing "or PA", that's something like a copy-paste error.

klown: @mbrambilla what kind of grief am I going to cause you due to a refactoring of the `popover_tour.html` template?

src/shared/templates/shared/partial/popover_tour.html
klown: The general rationale of `has_teacher_resource` was flexibility in showing the link.  It should have been named `has_teacher_resource_link` or, even better, `show_teacher_resource_link`.

Admittedly it's somewhat muddled in that the link is not shown for student and guest users, somthing that could be handled by `show_teacher_resource_link`.  But the template covers that situation by the check for a TE or PA role*.

Still, there are cases where the user is a teacher or a parent and the link should not be shown.  For example, if the user has navigated to the Manage page and the Manage popover shows, the link in the popover is odd since it just navigates to the page the user is viewing.  It should not be shown in this case.

As for the Library case, that's a bug; it should not be false.

\*Perhaps this is part of the refactoring:  Move the role check logic to `views.py` to set  `show_teacher_resource_link` and then use only that boolean in the template.  It would remove some clutter.

src/shared/templates/shared/partial/popover_tour.html
mbrambilla: @klown Most of my refactoring in popover_tour.html was around the 'footer' section, as seen here -  you will need to click the 'load diff' link to see the changes: [#336 - popover_tour.html](https://github.com/cast-org/clusive/pull/336/files#diff-eb5c13490b9c5999f2faa4709aacae3eb06246c6dce38c67a9f011f8533a0d3a)

The goal was to create a re-usable template for the footer section to reduce redundant template markup.

klown: Thanks @mbrambilla.  I thought about merging just your footer change into this PR, but it has features that don't apply here or might break things.  I'll work on making a suitable version.  Since that is where most of your refactoring was done, it should make merge conflicts smaller.
🤞

klown:
@boris, @mbrambrilla, I have made changes in response to the review comments, mainly to refactor and reorganize the `popover_tour.html` template.  I think it could be simplified further if some `context` such as the title of the popover and the video link were set in the `views.py`.  But I'm not sure what belongs in the template and what belongs in the code; and am leaning towards the titles being in the template.  On the other hand, if they and other `context` were in the code, then the large `{% if %}` could be eliminated.  Or, perhaps just moved into the `views.py` code.
```
{% if tour_name == ...%} 
... 
{% elif tour_name == ... %}
...
{ % elif ... %}
...
{% endif %}
```

In addition, I noticed another page that was not included in the show-tip logic.  The page is the `ResourcePage`.  The  result is that when using the Reader to view a teacher resource, there are no popovers, and, by implication, there will be no tour.  I modified the code to treat the `ResourcePage` as having the same tip types as the `Reading` page.  Does that make sense?

klown:
Regarding more refactoring, I think there is a structure for each popover along the following lines:
```
tour_properties: {
    'settings': {
        'title': 'Settings',
        'video': 'https://videos.sproutvideo.com/embed/a79edbb41414e7c72e/96f3dfacf85e2603?type=hd',
        'tip_text': '<p>Change the font, colors, and more! Settings let you choose the look, supports, and reading tools to make Clusive all your own.</p>',
        'resource_id': 'SSETTINGS',
        'resource_name': 'settings',
    },
    'readaloud': {
        'title': 'Read-aloud',
        'video': 'https://videos.sproutvideo.com/embed/449edbb51c1ae6cecd/90717ea3b61bbdd3?type=hd',
        'tip_text': '<p>Click the page read aloud icon to hear the entire page. Select text and click read aloud in the context menu to hear a smaller selection or a single word read aloud.</p>',
        'resource_id': None,
        'resource_name': 'None,
    },
    ...
}
```
It's a lookup-table where each tip type (e.g., `settings`) defines a set of values for that popover dialog, namely, its title, the link to its video, the tip text, and its teacher resource link, if any.  The properties appropriate to the response are keyed by the `tip_shown` in the `context` and are set as `tour_properties`.

In that case, I _think_ the template's `{%if tour_name == ...%) ... {% elif %} ... {% endif %}` is reduced to just one template where the `tour_properties` fields fill at the appropriate places.

klown:
I wrote:
> Regarding more refactoring, I think there is a structure for each popover along the following lines:

I pushed a version of this approach to [another branch](https://github.com/cast-org/clusive/tree/feature/CSL-2040-group-2-enhanced-tooltips-refactor) based on this one.  [This](https://github.com/cast-org/clusive/blob/feature/CSL-2040-group-2-enhanced-tooltips-refactor/src/shared/templates/shared/partial/popover_tour.html#L29) shows the markup for all but one of the popover dialogs.  The `context` popover turns out to be a special case.

PR #327, opened 2022-10-13
CSL-2007: Additional support for eventual chaining

mbrambilla:
Pushing a a PR to a PR  to hopefully not blow up work by @klown 

Removed 'robust' flag since this is always true for the current set of popovers. To test the example chain of 2 items
- uncomment the array of names in `src/pages/views.py` at line 728
- go to any publication and in developer console run the command `tour.chain('#tour_settings');` to start the tour
- actual start button has not been added yet as it is still in design

Other changes:
- includes some new template filters to help with the array parsing
- moved the for loop into the `popover_tour.html` template
- moved the 'footer' navigation buttons and learn more links into their own template `popover_tour_footer.html`

PR #333, opened 2022-10-24
Feature/csl 2041 popover tour

mbrambilla:
Dependency: #328
Supersedes: #327

Re-implements the chained popover tour based on revised popover tips.

The 'singleton' tour item, or the one that shows up automatically based on page and history conditions needs to be part of the larger tour list for a given page.  For example, on the reader page in order to show `'settings'` popover on page load it needs to be included in the full tour list - currently comprised of the list `['switch', 'settings', 'readaloud', 'context', 'thoughts', 'wordbank']`

### DIFF
--- a/src/frontend/js/frontend.js
+++ b/src/frontend/js/frontend.js
@@ -1298,6 +1298,34 @@ function setUpDeferredLoadOfCompDetails() {
     });
 }
 
+function dialogMediaStop() {
+    'use strict';
+
+    function dialogTarget(evt) {
+        if (evt.namespace.includes('popover')) {
+            return $(evt.target).data('cfw.popover').$target[0];
+        }
+        if (evt.namespace.includes('modal')) {
+            return $(evt.target).data('cfw.modal').$target[0];
+        }
+        return null;
+    }
+
+    $(document.body).on('beforeHide.cfw.popover beforeHide.cfw.modal', function(evt) {
+        var target = dialogTarget(evt);
+        if (target === null) { return; }
+
+        var $iframes = $(target).find('iframe[src]');
+        $iframes.each(function() {
+            var source = this.src;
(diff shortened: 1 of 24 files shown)


mbrambilla:
- 2 second delay on show has been added for the singleton popovers, the ones shown on page load, to match the tooltip delay
- Some adjustments to improve the visuals due to tooltip and popover animations.
- Youtube and SproutVideo `<iframe>`s will have their `src` attribute reset when a dialog is closed - deals with the issue where media continuing to play when dialog is closed and of works around having to implement API interfaces for either video service.

@klown  @bgoldowsky  The Cluey, or more properly named 'tour', tooltip has been stubbed out for the Dashboard, but I am not sure what the proper method for adding it to the `tiptypes.json` file, as I think the priorities may need to be revised so that the 'tour' tooltip comes up before the popover singletons.

klown:
@mbrambilla The Cluey tip was in [`tiptypes.json`](https://github.com/cast-org/clusive/blob/a85adc22dcf0fcd68b4c078317a79e1479ad1abf/src/tips/fixtures/tiptypes.json#L113) in a previous revision, and in the `TipType` database table with priority 12, but then it was removed when its enhanced tooltip was removed.  It can be put back, but what is its priority?  Is it "1" -- all things being equal, it should be the first popup/popover singleton that shows up?  Currently, the tip type with priority 1 is `settings`, and that was set up from when the tips were presented as plain tool tips.

If it is "1" and all the other types are increased by "1", then maybe a migration would adjust the `TipType` table in the database?  Running `loaddata` results in unique constraint errors on the `priority` field.  And, flushing the TipType table prior to `loaddata` incurs a plethora of protected Events that need to be deleted first.

mbrambilla:
In the process of re-implementing this.  My local merge went horribly wrong somewhere.  New PR inbound.

PR #336, opened 2022-10-28
CSL-2041: Popover tours

mbrambilla:
Re-implementation of PR #333.

As noted previously, there are updates to the `tips_tiptype` that are needed.  Please refer to notes from @klown in [PR #](https://github.com/cast-org/clusive/pull/333#issuecomment-1291090083) regarding `[tiptypes.json](https://github.com/cast-org/clusive/blob/0e70bc499b1566f8fe00afe77ff1ef2330b4517d/src/tips/fixtures/tiptypes.json)` and issues with updating the db. 

I am are there is a way to get a migration file, as this is also possibly needed for PR #328. ~~Potential solution: https://stackoverflow.com/questions/4002322/migrating-django-fixtures~~ Actually this appears to be more of a create fixture, not a migration.

Reference: https://code.djangoproject.com/ticket/24778


klown:
I don't believe any migration is required for PR #328 because the database _structure_ for the `TipType`s remains the same.  That is, no new columns are needed.

What was needed was to add a number of records for the new types.  The last priority value before adding the new set was `8` for the `activity` tip.  I added the new tip types starting with priority `9` for `student_reactions`, increasing by `1` for every new tip.  Running `loaddata` gave no errors.

However, when later removing the `cluey` tip, it created a hole and the priorities needed to shift.  Running `loadata` thereafter resulted in uniqueness errors.  I solved that by hand:  Using the administrator's interface to the database, I removed all of the tip types and all related `Events`.  It wasn't too bad since it's a development site and I often clear `Events` out anyway, just to save space.  Thereafter, I could load the new tip types and their latest priorities from the json fixture into an empty `TipType` table.

But that approach is not feasible for Q/A nor Production since there are a lot of Event records there, and I assume we want to keep them, and continue to have them reference the same tip types as they do now.

I'll look at the stack overflow article and also search for any documents regarding migration and unique fields.  Another possibility is to write a python script that:

1. Get the tip with the largest priority,
2. Increase its priority by one and save it,
3. Repeat until all current tips have a new priority one greater than before.

That should create a hole at the top that allows `loaddata` to run without a uniqueness error and add a `tour` tip type with priority `1`.

klown:
@mbrambilla, I put together a quick and dirty `incrementpriorities` that does what I outlined above to increment each tip type's priority by one, starting with the tip that has the highest priority.  It works and creates a "hole" in that no tip type has a priority of `1` any longer.  Running `loaddata` thereafter where the `tiptypes.json` has a `tour` type with a priority value equal to `1` succeeds.
 
But, I noticed that all of the tip type's primary keys initially matched their priority.  After running the increment manage command, they no longer match.  I don't know if that's important.  Also, I'm almost certain that the primary keys cannot be changed since that is what the database actually uses to associate the tip type with other records like `Event`s, or `TipHistory`s.  In addition, the json fixture file defines the primary key with the `'pk'` field for each type.  The new `tour` tip had to have a `'pk': 16` to make sure its primary key didn't collide with any other type.

I have to ponder this some more.

klown:
@mbrambilla If you want to check out the script, I've created a draft branch and PR based on this branch/PR in PR #337.

Use with caution :-)

bgoldowsky:
The priority can certainly be different from the pk - that they matched is just historical accident.  Pk's should not be changed to avoid breaking DB connections, but priorities can be changed as needed.  As far as I recall, the actual value of priority also doesn't matter - just the ordering. So holes are fine.

klown:
Thanks for the clarifications @bgoldowsky !

In that case, is the `incrementpriorities` manage command good enough?  It creates a hole in terms of priorities such that the `tour` tip can be inserted.  That command is part of PR #337, as well as a version of `tiptypes.json` that includes the `tour` tip with a priority of `1`.

bgoldowsky:
@mbrambilla @klown I made some changes, would love to get your feedback on them.

There is one remaining issue; there's at least one case where a tip is legal but just can't attach to anything (when a brand new user doesn't have a react wheel on their dashboard yet because they haven't read a book).  This type of case is handled reasonably gracefully for single tips (that's why we added the 'tip viewed' message sent back from the client side), but breaks the tour with a JS error.  We either need to (a) put some defensive code in the tour JS that will skip over tips that are not actually anchored to the page, or (b) add more conditions to `TipType.can_show()` to try to catch this and any other case where a tooltip won't be able to be shown.  WDYT?

mbrambilla:
Some checking is possible in the `tour.chain()` JS side, but then the tour list would need to be passed to the JS and the popover footers would need to be dynamically updated if an item is missing or hidden.

The current implementation has the inter-linking of the tour popovers is handled through the template code through the loop iteration (helped by the `list_index_prev` and `list_index_next` filters) so we didn't have to do all the chaining, counting, and HTML manipulation through JS.

We might also need to handle the possible case of items being shown/hidden based on the tab interface of some of the panels (I never double-checked).

However, if it is this one case, would it might be easier (faster) to remove the missing tour item from the list on the Django side based on a conditional to see if the student has not read any books.

src/tips/fixtures/tiptypes.json
```diff
@@ -154,7 +154,7 @@
     "pk": 16,
     "fields": {
       "name": "tour",
-      "priority": 1,
+      "priority": 0,
```
klown, line 172: Very nice.  And, it makes PR #337 to shuffle the priorities around unnecessary*.

\<curmudgeon>And here I thought that zero was neither positive nor negative.  Django's `PositiveSmallIntegerField` says otherwise\</curmudgeon>

\* for now...

bgoldowsky: One of those areas where math ≠ computer science, I guess?

bgoldowsky:
Why is 'thoughts' classified as student only?  Any reason not to include it in the teacher tour of the reading page?

bgoldowsky:
BTW I am working on revising can_show() along the lines of Matt's suggestion.

klown:
I'm not for changing `TipType.can_show()` since it's already relatively complicated.  However, @mbrambilla 's outline of how complicated it is to handle missing tour popovers in the JS suggests the `can_show()` is the lesser of two evils.

Regarding the missing popover "breaks the tour with a JS error": what is the nature of the error?  Is it an exception that could be caught and somehow skipped over?  Or some generic error?

bgoldowsky:
It's a NPE because it tries to activate a popover in the chain but it's not there. The error could be caught, but would be awkward at best - the number of steps in the tour shown in the footer would be wrong, 'next' or 'prev' buttons would be active when they shouldn't be, etc. I think Matt is right that we need to make the Django side accurate as to what items are going to be available.  I think I can make it not too ugly.

klown:
> Why is 'thoughts' classified as student only? Any reason not to include it in the teacher tour of the reading page?

That's a question for Kristin/Kim/Agatha.  Kristin listed it as such in [CSL-2040](https://castudl.atlassian.net/browse/CSL-2040?focusedCommentId=36756), and the [content](https://docs.google.com/document/d/1cO4fBWJWONSPIAqR0HGPwvC8B9iPNQCdb3XErmV8RLs/edit#heading=h.hl1e3ma03e1q) (google doc) also shows it as "student only" under the "Link to video [teacher]" heading: "[student only tip]".

bgoldowsky:
I think that was probably a misunderstanding since there was considerable confusion between 'thoughts' and 'student_reactions'. I can check with Kristin.

klown:
More re: "thoughts" and it being student only.  I think there's a bug here.  The video for student-only "thoughts" is actually for the student_reactions tip on the student dashboard.  There was a time when that tip for the student dashboard was called "thoughts" which raised other problems.  I switched it to student_reactions for the student dashboard, but I may not have updated `can_show()` to reflect that.  I'm still investigating and I believe that looking at the popover's contents for the "Your thoughts" widget on the reading page would help to decide, but I'm having some difficulty finding it.

klown:
I found it and it is aimed at students.  Part of the text for the popover notes that all of your reactions can be found on the student dashboard.  The video is somewhat less student-centric, but at one point says, "... your teacher ..." which implies the user is a student, and also points out the reaction wheel on the student dashboard.  Still the beginning of the video is fairly generic in describing the dialog for filling in reactions, answering the custom question, and so on.

Here's the link:  https://docs.google.com/document/d/1SYru98tt_Zmzz2z_7lKDCSiKqsOdDFTTLp0XMvW97h0/edit
Note that the document says "Student only tip" here as well.

So, I take back that I think there is bug in `can_show()` :-)

bgoldowsky:
Thanks @klown for the investigation. I acknowledge that Kristin has specified 'thoughts' as student-only.  Nevertheless, I am going to go with my gut feeling for now that there is no reason for it to be the single outlier student-only tooltip, until I can actually have a conversation with Kristin about it.

PR #343, opened 2022-11-11
Feature/CSL-2046 v2 student detail page

klown:
Branch for tracking all the student details page work, including its sub-tasks.

JIRA: https://castudl.atlassian.net/browse/CSL-2046


src/roster/templates/roster/student-details.html
```diff
@@ -0,0 +1,203 @@
...
+    </div>
+</main>
+{# Unsure if the tour or tips are available on the student detail page #}
+{# Leaving in place for now #}
+{% for tour in tours %}
+{% include "shared/partial/popover_tour.html" with tour_name=tour.name tour_robust=tour.robust %}
+{% endfor %}
+{% if tip_name %}
+{% include "shared/partial/tooltip_tip.html" %}
+{% endif %}
```
mbrambilla: Not needed as this is already included in the [base.html](https://github.com/cast-org/clusive/blob/40837425056fcf1f65420ddb4067280031730b76/src/shared/templates/shared/base.html#L74-L79) template.  This template builds on top of the base.

klown: Thanks @mbrambilla ; removed.  But, I've still not added the `libraryMasonryEnable();` script until the layout issue with the template and FF is worked out.

@cindyli this is now ready to merge with your pull requests to get the correct styles.

mbrambilla: @klown Just double checked with MacOS and Win FF with the function call - the Masonry layout is working fine for me.  Chrome, Safari, and Edge are also working.  Checked with reduced motion/animation OS settings also.

Could it be a browser cache issue?

klown: @mbrambilla  Just now I added the script to the template for the first time today, and the layout of the details page is fine in FF.  I've pushed the change here.

@cindyli The student details template has one more change.  Could you merge it with your work and see if the layout is okay for you, too?  Thanks.

cindyli: @klown Merged this branch into my working branch. Tested with FF, Chrome and Safari on Mac. The layout looks fine on them.

klown: 🎊

src/roster/views.py
```diff
@@ -1466,3 +1467,64 @@ def remove_social_account(request, *args, **kwargs):
...
+        if 'days' in kwargs:
+            self.days = kwargs.get('days')
+            logger.debug('Setting student activity days = %d', self.days)
+            self.clusive_user.student_activity_days = self.days
+            self.clusive_user.save()
+        else:
+            self.days = self.clusive_user.student_activity_days
+
+        period = self.clusive_user.current_period
+        self.roster = period.users.exclude(user=request.user, role=Roles.TEACHER).order_by('user__first_name')
```
bgoldowsky: This should look for STUDENT role.  Currently it incorrectly includes parents in the displayed info.

src/roster/views.py
```diff
@@ -1466,3 +1467,64 @@ def remove_social_account(request, *args, **kwargs):
...
+        # Dictionary for the individual panel data
+        self.panel_data = dict()
+        self.panel_data['days'] = self.days
+        try:
+            self.clusive_student = ClusiveUser.objects.get(
+                user__username=kwargs['username'],
+                periods__in=[period],
+                role__in=[Roles.STUDENT, Roles.GUEST]
```
bgoldowsky: Guest not needed - would never be a member of a period.

bgoldowsky (commented):
src/roster/templates/roster/student-details.html should be renamed with an underscore instead of a hyphen, to match all the other templates.

klown:
> src/roster/templates/roster/student-details.html should be renamed with an underscore instead of a hyphen, to match all the other templates.

It looks like I copied the name from the mockup:  `src/frontend/html/student-details.html`

I'll use `git mv` to change the name.

src/roster/views.py
klown: Makes sense.

src/roster/views.py
klown: Good catch.  I should add more parent users since, as it is now, I won't see any parents when I test this.

src/roster/views.py
klown: Note to self, change to:
```
self.clusive_student = ClusiveUser.objects.get(
    user__username=kwargs['username'],
    periods__in=[period],
    role=Roles.STUDENT
)
```

bgoldowsky:
I have more testing to do but a couple of comments from a little poking around:
* We should think about empty state(s).
* If there is only one version of a reading, should not show "Challenge 1" in the last column.
* A line break between the hours and visits in the second column would improve the layout.
* Open responses to the standard comp question are not showing up in the "learning" column, just the nothing/a little/a lot rating. Are they supposed to?
* If a book is customized but without any custom question, "()" shows up instead of "—" in the 6th column.

klown:
Aside, from the empty states issue, these are all relative to the reading details table/panel.  I'll start working on them.
> 
> * If there is only one version of a reading, should not show "Challenge 1" in the last column.
> * A line break between the hours and visits in the second column would improve the layout.
> * Open responses to the standard comp question are not showing up in the "learning" column, just the nothing/a little/a lot rating. Are they supposed to?
> * If a book is customized but without any custom question, "()" shows up instead of "—" in the 6th column.

In addition, I've noticed that in one case some of the links in the student affect wheel lead to the modal saying, for example, "No readings have made Samuel feel okay yet." even though the "okay" slice is filled in. I'll look into that too.

klown:
@bgoldowsky Your issues as well as the way to display the date for the "Last read" column have been addressed:

- Empty states handled as suggested by [Kim's comment](https://castudl.atlassian.net/browse/CSL-2046?focusedCommentId=37671) and [your response](https://castudl.atlassian.net/browse/CSL-2046?focusedCommentId=37676) in the JIRA,
- "Challenge" removed from "Reading level" column when there is only one version of the reading,
- Line break added between hours and visits in the "Active time" column,
- Free responses combined with ratings in the "Learning" column when a response is available,
- Show "—" in "Custom" column when there is no custom question.
- Dates in the "Last read" column displayed as "Day, Mon nn", e.g., "Fri, Nov 25" as per [Agatha's recommendation](https://castudl.atlassian.net/browse/CSL-2046?focusedCommentId=37697).

@mbrambilla regarding the display of last read date, I added a `style=white-space:nowrap';` to the `<td>` element to keep the date from wrapping.  I looked through the css files in the `frontend` folder, but I couldn't find a class that used that style without a lot of other styles.  Is there one?  Here's a reference to the use in the student's reading details table -- thanks for any advice here:
 https://github.com/cast-org/clusive/blob/37459b760655770d9c7e2c77608e19c6048e0ddc/src/roster/templates/roster/partial/student_details_reading_details.html#L70

klown:
I wrote:

> In addition, I've noticed that in one case some of the links in the student affect wheel lead to the modal saying, for example, "No readings have made Samuel feel okay yet." even though the "okay" slice is filled in. I'll look into that too.

I traced this back to the database and found that (1) there are `AffectiveUserTotal` records for the student with counts for certain affect words, but (2) no corresponding `AffectCheckResponse` records for that word.  Since the dialog that is presented when clicking on the reaction wheel is based on the `AffectCheckResponse` records found, it reports that no readings made the user react that way.

Since my local development database is not pristine, I suspect it's a case of GIGO, and that there is no real bug here.

klown:
I had another thought about `AffectiveUserTotal`s and `AffectCheckResponse`s.  The latter refer to a book, and if that book is deleted, the check response is also deleted.  The `AffectiveUserTotal`s are not affected, however, since they have no direct connection to any readings.  So, it is possible for a user's affect summaries to hold reactions but lose the context (the readings) of that reaction.

bgoldowsky:
> * "Challenge" removed from "Reading level" column when there is only one version of the reading,

I am not seeing challenge listed at all now, even when there are multiple versions.

bgoldowsky:
OK, so other than the small issue mentioned above, this is working very well. I do think we should do a round of efficiency improvements, however. On a modest sample user (who has about 15 readings), DJDT is showing 259 SQL queries were run.  A couple of thoughts on how to cut this down:

1. Use `prefetch_related` on some queries so that information is already present without having to do separate queries later. For example in `get_reading_data`, it queries for all the user's Paradatas, then there are separate queries that look up the Book for each Paradata and whether it was assigned.
2. Only call `get_book_details` for the 10 books that are going to be displayed on the current page, not all books they ever read.

klown:
@bgoldowsky wrote:
> I am not seeing challenge listed at all now, even when there are multiple versions.

The latest push should fix that.  I also added a line break between the reading levels and the challenge level, as shown in the wire-frames.

I'll start looking at the db access improvements you suggested.

klown:
@bgoldowsky, the latest push calls `get_book_details()` for just the books shown on the page.  It's noticeably faster when changing the sort and when paging to a given page.

However, there is some fiddling so that the paging control works correctly that I'm less satisfied with.  It's good enough for now.  I'll next look into `prefetch_related` next, but if I think of improvements for paging, I'll switch back to that.

PR #349, opened 2022-11-17
CSL-2107: Implement time scale selector on the student details page

cindyli:


cindyli:
Replaced by [the pull request](https://github.com/cast-org/clusive/pull/350) issued against the cast-org repository.

src/roster/views.py
```diff
@@ -1475,6 +1475,7 @@ def __init__(self):
 
     def get(self, request, *args, **kwargs):
         self.clusive_user = request.clusive_user
+        self.days = kwargs['days']
```
klown: We need a default value for `days` here because one of the endpoints is `/details/<username>` with no time scale specified.

Regarding defaults, @bgoldowsky suggested in the JIRA that "We can use the existing ClusiveUser field as the persisted default choice, so that whichever time period you have picked on the dashboard carries over to this page (may want to discuss if this is the desired UX)"

I agree with @bgoldowsky 's suggestion; intuitively, if the teacher has intentionally set a time scale on the dashboard, they probably expect the same time scale on the details page.  I would go with that until UX says otherwise.

Also, @cindyli suggested the other day (private conversation) that the students listed on the Manage page could also be links to their details page.  Currently they are just static text.  I agree with @cindyli's suggestion, but there is no time scale associated with the Manage page.  Overall, if we continue to support the `/details/<username>` endpoint, we need some default time scale.  The Dashboard defaults to 7 days (I think), so we could do the same here:
```
self.days = kwargs.get('days', 7)
```

cindyli: Since the requirement so far to land on the student details page is from the dashboard "Active reading time" panel where a time is always selected and will be carried to the student details page. I've removed the support of `/details/<username>`.

klown: 👍 
I thought about it a bit more.

The student links on the Dashboard are inside the "Active reading time".  That context implies that the user is requesting more details about that student and reading information.

Having links to student details on the Manage page lacks that context so it's not obvious where the links go in that case.  Also, the Manage page is about managing the period -- modifying the roster, modifying individual student account information, and the name of the period.  That seems distant from the kind of information the student details link would provide.

For that reason, I no longer think that details links are useful on the Manage page.  But, in any case, I have asked the UX team for their opinion.

klown:
While trying this out and comparing it to how the time frame is selected in the dashboard I noticed that the dashboard uses ajax:
- the ajax request is, for a month, `GET /dash-board-actitivy-panel/30`.  There is no body.
- the response is the markup for the entire "Active reading time" panel, starting with `<div id="StudentActivityPanel" class="col">...`

I'm not sure if this has implications for the way time scale selection works here for the student details page where it's a standard request for the entire page, but with the new time scale specified in the request url.  I think that's fine since the time scale affects a number of the panels on the page, and not just the reading activities panel.

So I'm satisfied that this PR can be merged with the parent student details branch, mostly for the other panels that need the time frame code.  However, if you want to take a look first before I merge it @bgoldowsky, let me know.

cindyli:
@klown and @bgoldowsky, as we discussed in today's tech meeting, the selection of the days on the time scale is now saved into the database.

This pull request is ready for more reviews. Thanks.

src/roster/views.py
```diff
@@ -1475,6 +1475,15 @@ def __init__(self):
...
+
+        if 'days' in kwargs:
+            self.days = kwargs.get('days')
+            logger.debug('Setting student activity days = %d', self.days)
+            self.clusive_user.student_activity_days = self.days
+            self.clusive_user.save()
+        else:
+            self.days = self.clusive_user.student_activity_days
+
         if not self.clusive_user.can_manage_periods:
```
klown, line 1487: I should have noticed this for the time scale PR:  I think this check whether the current user can manage periods should happen earlier, right after assigning `self.clusive_user`.  The reason is that there is no need to do anything further if the user cannot navigate to the student details page.  If the check fails, `self.handle_no_permission()` is called and results in a "403 Forbidden" response.

klown: Ignore the above -- it was meant for your Activies PR #350.  I'll put it there.

PR #350, opened 2022-11-17
CSL-2102: Implement the activity panel on the student details page

cindyli:
This pull request is dependent on [the pull request that implements the time scale](https://github.com/cast-org/clusive/pull/349), which should be reviewed and merged first.

The dependency is because the data on the activity panel needs to reflect the selection on the time scale.


src/roster/views.py
```diff
@@ -1495,6 +1497,23 @@ def get(self, request, *args, **kwargs):
         # TODO: (JS) fill these in as necessary.
         self.panels = dict()
         self.panel_data = dict()
+
+        # Student Activity panel
+        self.panels['activity'] = self.clusive_student
+        if self.panels['activity']:
+            students_reading_data = Paradata.reading_data_for_period(self.clusive_user.current_period, days=self.days, sort='name')
+            target_reading_data = {}
+            for one_reading_data in students_reading_data:
```
klown: I'm wondering if it's worth it to make this loop a class function of Paradata, something like:
```
@classmethod
def reading_data_for_student_period(student: ClusiveUser, period: Period, days, sort)
```

cindyli: I agree it would be helpful to have such a class function.

I'm thinking the code might be more efficient by adding an optional argument `username` to reading_data_for_period(). If `username == None`, look up for all students, otherwise, only look up for the given student.

The current function name may need to be adjusted to reflect this new feature.

Let me know what you think. Thanks.

src/roster/views.py
```diff
@@ -1475,6 +1476,15 @@ def __init__(self):
...
+
+        if 'days' in kwargs:
+            self.days = kwargs.get('days')
+            logger.debug('Setting student activity days = %d', self.days)
+            self.clusive_user.student_activity_days = self.days
+            self.clusive_user.save()
+        else:
+            self.days = self.clusive_user.student_activity_days
+
         if not self.clusive_user.can_manage_periods:
```
klown: I should have noticed this for the time scale PR: I think this check whether the current user can manage periods should happen earlier, right after assigning self.clusive_user. The reason is that there is no need to do anything further if the user cannot navigate to the student details page. If the check fails, self.handle_no_permission() is called and results in a "403 Forbidden" response.

bgoldowsky: Agreed. This check can go right at the top.

src/roster/views.py
bgoldowsky: Either way would be a big improvement over extracting the data for all students and then throwing away everything that's not the current student.

We don't want _everything_ that is in `reading_data_for_student_period` though. In particular, that method has code to collapse readings that are only a small amount of time into a single "other" record. We don't want to do that for the details page. We don't need to calculate the time percentages at all.

The most elegant would be split it into two methods - a new class method that pulls basic data either for a full class or a single student - and then change `reading_data_for_student_period` to call that and do its extra processing, and the student details view method to call that and _not_ do the extra processing.

src/roster/views.py
```diff
@@ -1504,5 +1533,7 @@ def get_context_data(self, **kwargs):
         context['current_student_name'] = self.clusive_student.user.first_name if self.clusive_student else "No student"
         context['teacher'] = self.clusive_user
         context['roster'] = self.roster
-        context['data'] = { 'days': 7 }
+        context['data'] = { 'days': self.days }
```
klown: My approach was to follow the Dashboard, putting all of the panel information in `context['data']`.  One change to the Dashboard's approach was to call the python dictionary `self.panel_data` instead of `self.data`, mostly because I kept forgetting what `self.data` was for.

When I merged the time scale selection code into my student reactions branch, I made a change:
```
self.panel_data['days'] = self.days
...
self.panel_data['affect'] = { ... }
...
context['data'] = self.panel_data
```
There are two ways forward.  One is to put all of the panel data into `context['panel_data']` and get rid of `context['data']`, or put all of the panel data into `context['data']`, and get rid of `context['panel_data']`.

I lean towards the first solution (`context[panel_data]`), since it's clearer what the data is for.  However, it might not be possible since some of the included templates use `data` variables, and they are shared with other parts of the system.

It's likely easier/safer to switch to putting everything into `context['data']` and use `{{ data }}` in your templates.

src/roster/views.py
bgoldowsky: Either way would be a big improvement over extracting the data for all students and then throwing away everything that's not the current student.

We don't want _everything_ that is in `reading_data_for_student_period` though. In particular, that method has code to collapse readings that are only a small amount of time into a single "other" record. We don't want to do that for the details page. We don't need to calculate the time percentages at all.

The most elegant would be split it into two methods - a new class method that pulls basic data either for a full class or a single student - and then change `reading_data_for_student_period` to call that and do its extra processing, and the student details view method to call that and _not_ do the extra processing.

src/roster/views.py
cindyli: Good point, @klown. I implemented the suggestion of using `context['panel_data']` for holding `self.days`.

cindyli:
@klown and @bgoldowsky, thanks for the code review. All comments have been addressed. This pull request is ready for another review.

src/library/models.py
```diff
@@ -786,7 +807,7 @@ def update_id(self):
         so that client & server agree on one ID."""
         hl = self.highlight_object
         hl['id'] = self.pk
-        self.highlight_object = hl
+        highlight_object = hl
```
bgoldowsky: This change seems unrelated to the purpose of the PR and potentially problematic.

bgoldowsky (commented):
The documentation is clear, but the method names `reading_books_for_period_student` and `reading_data_for_period_student` don't really give me much clue about what the difference is.  Maybe something like `get_reading_data` and `get_truncated_reading_data` ?

Other than that and one inline comment that might have been an accidental change, looks great.

src/roster/views.py
```diff
@@ -1490,11 +1500,26 @@ def get(self, request, *args, **kwargs):
             self.clusive_student = None
 
         self.roster = period.users.exclude(user=request.user, role=Roles.TEACHER).order_by('user__first_name')
+
+        # Get the reading data for the current student. This data will be shared by all panels on the student details page
+        reading_data = Paradata.reading_books_for_period_student(self.clusive_user.current_period, days=self.days, sort='name', username=kwargs['username'])[0]
```
klown: Minor: I suggest using `.first()` instead of `[0]`, especially if the set can be empty.  The `.first()` function will return `None` if it is empty, whereas `[0]` will raise an exception.

See:  https://docs.djangoproject.com/en/3.2/ref/models/querysets/#first

src/roster/views.py
```diff
@@ -1504,5 +1529,6 @@ def get_context_data(self, **kwargs):
         context['current_student_name'] = self.clusive_student.user.first_name if self.clusive_student else "No student"
         context['teacher'] = self.clusive_user
         context['roster'] = self.roster
-        context['data'] = { 'days': 7 }
+        context['panels'] = self.panels
```
klown: I didn't do a thorough check, but I do not see `context['panels']` used anywhere in the templates.  Is it used?

I copied it and `self.panels` from the dashboard code for the basic student details template.  I have since looked back at the dashboard and they are used there because which panels are shown depends on a number of factors. The dashboard template has a series of `{% if {{panels.activity %} ... {% endif %} ...` statements for checking which panels are present.

In the case of the student details page, all of the panels are present, so I _think_ it's no longer needed.  I've removed it from the student reactions panel code.

But if it is used in your code, then it stays.

src/roster/views.py
klown: LGTM 👍

klown (commented):
This looks great @cindyli !

It would be useful to run the unit tests to check that the model changes do not break anything.  Do the following in the target directory to run the tests:
```
$ python manage.py test
```
It will take about 30 seconds -- there are a lot of tests -- but if everything passes, the last line output will look like:
```
Ran 90 tests in 29.733s

OK
```
If something failed, the last line will look like:
```
Ran 90 tests in 29.733s

FAILED (failures=1)
```

If that happens, you can look through the output for lines that start `FAIL: ...` to see what broke.

src/roster/views.py
klown: Ignore my previous comment.  I thought that the return type was a`QuerySet` but it's a list, and `first()` won't work with that.

src/roster/views.py
cindyli: `context['panels']` is not used in my code too. I've removed related lines.

src/roster/views.py
cindyli: Your concern about no reading data being returned makes a lot of sense. I added the logic to handle this case.

src/library/models.py
cindyli: Good catch, @bgoldowsky. I've fixed this accidental change and run `python manage.py test` to make sure all tests pass.

cindyli:
@klown, @bgoldowsky, all new comments have been addressed. Ready for more reviews. Thanks.

bgoldowsky:
Looking good.  Only one comment, which is because I was not clear.  In my proposed names `get_truncated_reading_data` the "truncation" I was referring to was collapsing the potentially-long list of books into a shorter list with an 'other' category at the end.  So the names are switched from what you have.  I'm also completely open to better names for them.

Other than that one comment - I don't need to review again, @klown you can merge into your branch at any point.

cindyli:
> In my proposed names `get_truncated_reading_data` the "truncation" I was referring to was collapsing the potentially-long list of books into a shorter list with an 'other' category at the end. So the names are switched from what you have. I'm also completely open to better names for them.
> 
@bgoldowsky thanks for clarifying what the truncation means. Method names are now swapped.

src/roster/views.py
klown: LGTM

src/roster/views.py
klown: 👍

klown:
@bgoldowsky wrote:

>  I don't need to review again, @klown you can merge into your branch at any point.

Merged -- thanks @cindyli !

PR #351, opened 2022-11-18
CSL-2105 "Reactions" panel for v2 student details

klown:
- [x] implement [wire frame]( https://castudl.atlassian.net/browse/CSL-1847?focusedCommentId=36931) for modal when clicking an affect word link in the affect wheel:
  - in progress.
- [x] respond to time frame changes

JIRA: https://castudl.atlassian.net/browse/CSL-2105


klown:
Ready for review @bgoldowsky , with one quirky edge case.

If the URL specifies a student who is not in the class, the page show a message "Student '<username>' not in this class '<period name>', and the student popup selector shows "No student". 

With respect to the reactions panel itself:
- the affect wheel is empty since the student has no affect data for this class.  This seems correct to me.
- the affect word links work, however, and bring up the teacher/class modal as they would on the teacher dashboard's "Your students' reactions" panel.  This is the quirky bit.

The affect wheel could be rendered inert with some effort, but I don't know if it's worth it.  Having it interactive and showing some class information about the affect words doesn't feel completely wrong, just odd.  And, in that case, maybe the wedges should show the class information, but I'm not sure.

Let me know what you think.

klown:
I decided on a compromise regarding the `None` student case.  I've added text to the top of the panel explaining that the given student is not in this period, but here are all of the students' reactions for the class.  The rationale is that the teacher can then view the summary of, say, "happy" to see which student(s) they want to look closer at.  Then they can choose that student from the popup selection.

src/assessment/views.py
```diff
@@ -157,10 +157,22 @@ class AffectDetailView(LoginRequiredMixin, TemplateView):
...
 
     def get(self, request, *args, **kwargs):
         self.word = kwargs['word']
-        clusive_user = request.clusive_user
+
+        # If a username is provided by kwargs['for_user'], set the clusive_user
+        # to that user. If there is no such user, or no for_user in kwargs, use
+        # the request.clusive_user
+        self.for_user = kwargs.get('for_user')
+        try:
```
bgoldowsky: I wonder if, in the None case, this will actually query the database for ClusiveUser "where username is null".  Would be cleaner to handle that case with an if.

bgoldowsky (commented):
For consistentcy, can we rename the templates with hyphens to use underscores?

src/assessment/views.py
klown: Okay, done.

bgoldowsky:
I fixed up a couple of things, think this is good to merge into main student details branch.

PR #353, opened 2022-11-21
CSL-2103: Implement the topics panel on the student details page

cindyli:
This pull request is dependent on other 2 pull requests: https://github.com/cast-org/clusive/pull/349 and https://github.com/cast-org/clusive/pull/350 which should be reviewed and merged first.


klown (dismissed):
This looks good to me.  Passing over to @bgoldowsky for his opinion

src/roster/views.py
```diff
@@ -1512,10 +1512,23 @@ def get(self, request, *args, **kwargs):
...
                 'last_login': user.last_login
             }
+
+            # Topics panel
+            books = []
+            # Get all books for the current student
+            if reading_data:
+                for one_book in reading_data['books']:
+                    books.append(one_book['book_id'])
+            subjects = Subject.objects.filter(book__id__in=books).only('subject').values_list('subject', flat=True).distinct()
```
bgoldowsky, line 1522: I guess we should specify an order here.  Ideally, the order that the subjects are listed in the filter panel.

cindyli: @bgoldowsky is there a filter panel on the student details page? Do you mean the reading details panel? If it is, this is my understanding of your suggestion: every time when a user re-sort books in the reading panel, the order of topics in this panel should be adjusted accordingly with the new order of books? 

For now all topics in the topics panel are sorted in the default sorting order defined in the `Subject` model. Would it be confusing that topics shift around every time when a user sorts books in the reading panel?

I have a feeling that I misunderstood your suggestion. Please do correct me. Thanks.

cindyli: @bgoldowsky after thinking more of your comment, I think you probably meant topics should be displayed in the order of how they are displayed in the book description, which is red circled in the screen shot below.

Yes, the current implementation has taken it into consideration. The logic collects all books and query all their subjects in one shot. This query hits the default sort ordering defined in the `Meta` class of the `Subject` model, which will nicely return the desired order: "Fiction" and "Non-fiction" come before other subjects that are sorted in the alphabetical order.

Screen Shot 2022-12-02 at 11 11 01 AM

bgoldowsky: OK super.  Yes, that is what I meant. It is the same order as the filter panel that is on the library page. Since as you say it is set up as a default, you shouldn't need to change anything.

bgoldowsky:
Good to go.

PR #358, opened 2022-12-01
CSL-2101: Implement the reading details panel on the student details page

cindyli:
This pull request adds a new field `reading_details_sort` into the `ClusiveUser` model. When testing this PR, run `python manage.py migrate` in the target directory to migrate the database first.

TODOs:

1. Three sortable columns "title", "active time", "last read" on the reading details are `<a>` links by default. Once they are clicked and the column is sorted, they remain as links. While on the dashboard, student details panel, once a sortable column is sorted, the corresponding link becomes plain texts and cannot click again. Someone may need to verify with the design team if the reading panel should work the same as the dashboard.
2. The template for the pagination control is copied from the library page, so some css classes is namespaced with `library-`, for example at [these lines](https://github.com/cindyli/clusive/blob/feature/CSL-2101-reading-details/src/roster/templates/roster/partial/student_details_reading_details.html#L69-L70). If the pagination design is fine to be shared across various components, it might be less confusing to remove the `library-` namespace.


bgoldowsky:
It would be better for the currently-selected header to be a <span>, not a link. Links that do nothing are discouraged.

src/library/models.py
```diff
@@ -705,17 +711,27 @@ def get_reading_data(cls, period: Period, days=0, sort='name', username=None):
             result.sort(reverse=True, key=lambda item: item['book_count'])
         elif sort == StudentActivitySort.TIME:
             result.sort(reverse=True, key=lambda item: item['hours'])
+        
+        if (books_sort):
+            for reading_data_for_one_user in result:
+                if books_sort == ReadingDetailsSort.TITLE:
+                    reading_data_for_one_user['books'].sort(key=lambda item: item['title'])
```
bgoldowsky: Better would be `sort_title` rather than `title`.

src/roster/templates/roster/partial/student_details_reading_details.html
```diff
@@ -0,0 +1,104 @@
...
+                        <th class="d-sm-down-none">Learning</th>
+                        <th class="d-md-down-none">Custom</th>
+                        <th class="d-md-down-none">Reading level</th>
+                    </tr>
+                </thead>
+                <tbody>
+                    {% for book in data.page_obj %}
+                    <tr>
+                        <th scope="row">
+                            <a href="{% url 'reader_default' book_id=book.book_id %}">{{ book.title }}</a>
```
bgoldowsky, line 63: We will need to handle Bookshare books that the teacher is not able to view. Can borrow this code & design from dashboard.

src/roster/views.py
```diff
@@ -1468,7 +1470,91 @@ def remove_social_account(request, *args, **kwargs):
...
+        else:
+            self.sort = self.clusive_user.reading_details_sort
+
+        self.paginator = {}
+        self.page_obj = {}
+        try:
+            clusive_student = ClusiveUser.objects.get(
+                user__username=self.username,
+                periods__in=[self.current_period],
+                role__in=[Roles.STUDENT, Roles.GUEST]
```
bgoldowsky: Guest not needed

src/library/models.py
```diff
@@ -651,13 +651,14 @@ def latest_for_user(cls, user: ClusiveUser, max=10):
         return Paradata.objects.filter(user=user, last_view__isnull=False).order_by('-last_view')
 
     @classmethod
-    def get_reading_data(cls, period: Period, days=0, sort='name', username=None):
+    def get_reading_data(cls, period: Period, days=0, sort='name', books_sort='title', username=None):
```
bgoldowsky, line 676: I'm worried that the change to the default sort of books here will affect how they are displayed on the dashboard panel that also uses this method.  If so, should be easily fixed by passing in the correct sort from the dashboard code.

src/library/models.py
cindyli: @bgoldowsky could you clarify which `title` would be better to be `sort_title`? Do you mean change `ReadingDetailsSort.TITLE` to `ReadingDetailsSort.SORTTITLE`? I don't think the `title` in the variable `item['title']` can be changed as it's the field name defined in the `Book` model. Thanks.

src/library/models.py
cindyli: @bgoldowsky the dashboard doesn't offer the sorting function for individual books read by a user. Books on the dashboard are displayed in a random order at this point. The sort on the dashboard is at the user level on the total reading time of all books and total number of books that have been read. With adding the default sort of books, books on the dashboard will now be displayed in an alphabetical order of books.

If you are concerned about sorting books on book titles by default, I could change to the default to `None`. Please let me know what you think. Thanks.

src/library/models.py
bgoldowsky: Sorry. `Book` has both fields - I meant it would be better to sort on `Book.sort_title` rather than `Book.title`.

src/library/models.py
bgoldowsky: Current dashboard doesn't offer a choice of sorts, but it does include code to sort the books in descending order of time spent.  From development branch, library/models.py, line 702:
   `# Sort book entries by time`
   `entry['books'].sort(reverse=True, key=lambda p: p['hours'])`

cindyli: @bgoldowsky, thanks for pointing out the line. If sorting by hours is the default sorting order on the dashboard, and it's a duplicate with the new `book_sort` functionality. Do you think if it makes sense to set the default `book_sort` value to `time` instead of `title`? The line you pointed out can then be removed?

src/library/models.py
cindyli: Got it. Will make the change. Thanks.

src/library/models.py
cindyli: @bgoldowsky after a chat with @klown, I have a better understanding of this comment. The default `books_sort` value remains `title` for `get_reading_data()` and `get_truncated_reading_data()`. But in the dashboard code, `books_sort="time"` is explicitly passed to `get_truncated_reading_data()`. Here's [an example of changed code](https://github.com/cindyli/clusive/blob/feature/CSL-2101-reading-details/src/pages/views.py#L213). The line 702 in the library/models.py - `entry['books'].sort(reverse=True, key=lambda p: p['hours'])` - is no longer needed.

Hope this addresses your concern. Please let me know if it doesn't. Thanks.

cindyli:
@bgoldowsky and @klown, all review suggestions have been addressed. This pull request is ready for another review. Thanks.

src/roster/views.py
klown: Fixed in base CSL-2046 student detail branch (PR #343) and merged herein.

src/roster/views.py
```diff
@@ -1468,7 +1470,93 @@ def remove_social_account(request, *args, **kwargs):
...
+                'last_version': one_book['last_version'],
+                'custom_question': '(' + customizations[0].question + ')' if customizations else None,
+                'custom_response': custom_response[0].custom_response if custom_response else None,
+                'learning': comp_checks[0].get_answer if comp_checks else None,
+                'reading_level': ', '.join(category_names),
+                'is_assigned': one_book['is_assigned'],
+                'version_switched': True if one_book['first_version'] and one_book['first_version'] != one_book['last_version'] else False,
+                'unauthorized': not book.is_visible_to(clusiveUser)
+            })
+    print('=== book_details: ', book_details)
```
klown: Was this `print()` for debugging?  If so, remove it.  If not, better to user the `logger` here.

src/roster/views.py
```diff
@@ -1525,7 +1609,7 @@ def get(self, request, *args, **kwargs):
             }
 
         except ClusiveUser.DoesNotExist:
-            messages.error(request, f"Student '{kwargs['username']}' not in this class ({period.name})")
+            messages.error(request, "Student '" + kwargs['username'] + "' not in this class (" + period.name + ")")
```
klown: Why the change from the `f-string` technique?

cindyli: It's because I didn't know there is a `f-string` technique. I thought `f` is a typo and removed it. Then the message was not printed out properly so I had to change it to regular string concatenation.

klown: Oh.  If you still don't know, f-strings work like back ticks in javascript.  For example, using the same variable names, in JS it would be:
```
console.error(`Student '${kwargs['username']}' not in this class (${period.name})`);
```

bgoldowsky:
I think this can be merged. Will be easier to do final review/testing on the complete page all put together.

PR #359, opened 2022-12-02
Improve popover tour show logic to handle cases where items are …

mbrambilla:
…not present/visible

### DIFF
--- a/src/shared/templates/shared/partial/sidebar_bottom_end.html
+++ b/src/shared/templates/shared/partial/sidebar_bottom_end.html
@@ -3,8 +3,8 @@
     {% if tours %}
     <button id="tourStart" type="button" class="btn btn-tool btn-mascot-tipped mt-0_75" onclick="tour.chainStart();"><span class="sr-only">Start tour of features</span></button>
     {% else %}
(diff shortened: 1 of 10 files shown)


mbrambilla:
Added `scrollIntoView()` to both popovers and tooltips to try and move the tip/dialog into view after it is shown.  Fixes the case where the tip/dialog may be offscreen, such as on mobile with the tour tooltip, but mobile is still poorly supported with these large popovers.

src/shared/templates/shared/partial/sidebar_bottom_end.html
```diff
@@ -5,8 +5,8 @@
             data-clusive-tip-action="tour"
             onclick="tour.chain('#tour_{{ tours|list_index:0 }}');"><span class="sr-only">Start tour of features</span></button>
```
bgoldowsky: Should this be tour.chainStart() ? Clicking the button as is just throws an error for me.

mbrambilla: Ah, merge error on my part, pushing a fix

bgoldowsky:
Thanks for the fix!
Working well on dashboard and library, but I am getting errors invoking the tour on Wordbank or Manage (as teacher).

PR #366, opened 2022-12-15
CSL-1787: Update alt-profanity-check to v1.1.1 and python to v3.8.16

klown:
@bgoldowsky I updated my local development to use python to 3.8.16.  After running the unit and cypress tests successfully, and smoke-testing Clusive, I tried the upgrade of `alt-profanity-check` to version 1.1.1.  The unit, cypress and smoke-test all pass.  In particular, the simplify transformation appears to work -- I tested a word, and a paragraph.

As discussed at yesterday's technical meeting the warning messages no longer appear when you first run the server.  There's probably more things to investigate, but it's a start.


bgoldowsky:
Works for me!

PR #367, opened 2022-12-16
CSL-2046 Student details page fixes from Q/A issues

klown:
Q/A issues fixed:
- Reaction panel text, just above the reaction wheel
- Content of the reaction modal dialog:
  - If the reaction's free response is empty, do not show "None",
  - The do-not-enter sign is prepended to titles that the teacher is not authorized to view,
  - If any book is unauthorized, the unauthorized legend is displayed at the bottom of the dialog.
- Added spaces around the "/" in the Learning column that separates the amount learned and the free response,
- Handles corrupt data specifically when the last viewed time is `None`.  In that case, the book(s) is/are listed last.

JIRA:  https://castudl.atlassian.net/browse/CSL-2046

[CSL-2046]: https://castudl.atlassian.net/browse/CSL-2046?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/library/models.py
+++ b/src/library/models.py
@@ -744,7 +744,14 @@ class Paradata(models.Model):
                 if books_sort == ReadingDetailsSort.TIME:
                     reading_data_for_one_user['books'].sort(reverse=True, key=lambda item: item['hours'])
                 if books_sort == ReadingDetailsSort.LASTVIEW:
-                    reading_data_for_one_user['books'].sort(reverse=True, key=lambda item: item['last_view'])
+                    book_list = reading_data_for_one_user['books']
+                    last_view_null = []
+                    for abook in book_list:
+                        if abook['last_view'] is None:
+                            last_view_null.append(abook)
+                            book_list.remove(abook)
+                    book_list.sort(reverse=True, key=lambda item: item['last_view'])
+                    book_list.extend(last_view_null)
 
         return result
 
--- a/src/library/models.py
+++ b/src/library/models.py
@@ -5,7 +5,7 @@ import math
 import os
 import textwrap
 from base64 import b64encode
-from datetime import timedelta, date
+from datetime import timedelta, date, datetime, MINYEAR
 from json import JSONDecodeError
 
 from django.core.files.storage import default_storage
@@ -17,6 +17,8 @@ from roster.models import ClusiveUser, Period, Roles, StudentActivitySort, Readi
 from .util import sort_words_by_frequency
 
 logger = logging.getLogger(__name__)
+
+VERY_LONG_TIME_AGO = timezone.make_aware(datetime(year=MINYEAR, month=1, day=1))
 
 def update_word_list(word_list, word):
     """
@@ -744,15 +746,7 @@ class Paradata(models.Model):
                 if books_sort == ReadingDetailsSort.TIME:
                     reading_data_for_one_user['books'].sort(reverse=True, key=lambda item: item['hours'])
                 if books_sort == ReadingDetailsSort.LASTVIEW:
-                    book_list = reading_data_for_one_user['books']
-                    last_view_null = []
-                    for abook in book_list:
-                        if abook['last_view'] is None:
-                            last_view_null.append(abook)
-                            book_list.remove(abook)
-                    book_list.sort(reverse=True, key=lambda item: item['last_view'])
-                    book_list.extend(last_view_null)
-
+                    reading_data_for_one_user['books'].sort(reverse=True, key=lambda item: item['last_view'] or VERY_LONG_TIME_AGO)
         return result
 
     @classmethod
--- a/src/roster/templates/roster/partial/student_details_reading_details.html
+++ b/src/roster/templates/roster/partial/student_details_reading_details.html
@@ -92,7 +92,7 @@
                             </td>
                         </tr>
                         {% endfor %}
-                    {% elif current_student %}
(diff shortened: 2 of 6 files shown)


klown:
@bgoldowsky this is ready for review.

The fix for the corrupted `last_view` times is a little brute force and I'll think some more on it.  In summary, I added a loop to stash-and-remove all of the items with null `last_view` times, sort the remaining using the existing lambda, and then append the stashed items to the end of th sorted list.  I did not use `.desc(nulls_last=True)` because I wasn't sure it was relevant anymore.  The technique is sufficient, if a bit ugly.

src/library/models.py
```diff
@@ -744,7 +744,14 @@ def get_reading_data(cls, period: Period, days=0, sort='name', books_sort='title
...
                 if books_sort == ReadingDetailsSort.LASTVIEW:
-                    reading_data_for_one_user['books'].sort(reverse=True, key=lambda item: item['last_view'])
+                    book_list = reading_data_for_one_user['books']
+                    last_view_null = []
+                    for abook in book_list:
+                        if abook['last_view'] is None:
+                            last_view_null.append(abook)
+                            book_list.remove(abook)
+                    book_list.sort(reverse=True, key=lambda item: item['last_view'])
+                    book_list.extend(last_view_null)
```
bgoldowsky: Sure, that works, or you could do something like `sort(key=lambda item: item['last_view'] or a_very_old_date)`

src/roster/templates/roster/partial/student_details_reading_details.html
```diff
@@ -67,9 +67,21 @@ <h4 class="h2 mb-0">Reading details</h4>
...
                             </th>
                             <td>{{ book.hours }} <span aria-hidden="true">hrs</span><br><span class="sr-only">hours</span> ({{ book.view_count }} visits)</td>
-                            <td style="white-space:nowrap;">{{ book.last_view | naturalday:"D, M d" }}</td>
+                            <td style="white-space:nowrap;">
+                                {% if book.last_view %}
+                                    {{ book.last_view | naturalday:"D, M d" }}
+                                {% else %}
+                                    Unknown
+                                {% endif %}
+                            </td>
```
bgoldowsky, line 72: It might work here to just say `{ book.last_view | naturalday:"D, M d" | default:"Unknown" }}

src/shared/templates/shared/partial/modal_student_details_affect.html
```diff
@@ -7,13 +7,25 @@
...
+            {% endif %}
+            {% if reaction.affect_free_response and reaction.affect_free_response.strip %}
+                <br><em>{{ reaction.affect_free_response }}</em>
+            {% endif %}
         </li>
     {% endfor %}
     </ul>
+    {% if any_unauthorized_book %}
+    <p class="form-text"><em>Readings with the symbol <span class="icon-block" title="Not authorized to view"></span> were uploaded by the student.</em></p>
+    {% endif %}
```
bgoldowsky, line 28: This looks great.  We want to also do the same thing on the main page, if this icon appears in the readings table.

bgoldowsky (commented):
Looking good, I added a couple of comments.

src/library/models.py
klown: That's similar to an approach I abandoned.  During the earlier creation of the map, from which the `book_list` is derived, set the `last_view` property to a year ago if the database record's value was null.  Then, the sort would just proceed as before, putting all the items from the fake year ago at the end.  The problem was that you could not distinguish "a year ago" from a legitimate date.  If we had a constant set to `VERY_OLD_DATE`, that could act as a marker.  I'll think some more on that.

src/roster/templates/roster/partial/student_details_reading_details.html
klown: It does!  That makes it cleaner.

src/shared/templates/shared/partial/modal_student_details_affect.html
klown: Okay.  The simplest way is to show it always, similar to how the "Assigned" legend is shown at the top.  Due to paging, some pages of results will have no assigned books listed, yet "Assigned" is always present.  The "Unauthorized" legend could behave the same way -- always shown.

klown: The latest push addresses your latest three comments, @bgoldowsky.

In addition, I did write a version of the code that supports a conditional presentation of the "Unauthorized" legend at the bottom of the reading details table.  However, I still lean towards it being displayed unconditionally, so that is the version I pushed.  I do have a copy of the conditional version and can swap it in if necessary.

klown: I wrote:

> In addition, I did write a version of the code that supports a conditional presentation of the "Unauthorized" legend at the bottom of the reading details table. However, I still lean towards it being displayed unconditionally, so that is the version I pushed. I do have a copy of the conditional version and can swap it in if necessary.

Rather than store that version locally on my machine, I've made another PR #369 that includes the code for showing the "unauthorized" legend only when an unauthorized book is shown in the table.  PR #369 is based on the branch for this PR.  Thus, the code is the same for the two PRs except for the conditional display of the legend.

